PR #27, opened 2019-03-21
Add subset of Greek language translation

dspinellis:
Note: The ISO 639-1 language code for Greek is EL.


fredlcore:
Thank you! Also changed from LANG_GR.h to LANG_EL.

PR #29, opened 2019-04-02
Dutch translation

liudger:
Not complete but most stuff on my "Wärmepumpe" is working


fredlcore:
Thanks!

fredlcore:
Hi, regarding the two translations where you were not sure about:
"Funk" is not "function" but "Wireless" (for wireless room controllers)
"Zusatzerzeuger" (which you translated as "Extra vermogen" is an extra heat source.
Maybe you can check/update the translations if you find the time...

PR #31, opened 2019-04-07
Update Translation Dutch

liudger:
There were some errors so I cleaned the translation file


fredlcore:
Thanks again :) - and sorry for the delay!

PR #32, opened 2019-05-03
Correct DHCP

BudBundi:
Move block to display the IP after the initiation WiFi or Ethernet,
correct Ethernet start to make DHCP possible


fredlcore:
Thanks!

PR #38, opened 2019-08-04
Added support for separate DNS server

simplexify:
Added support for a DNS server which is different from the subnet's gateway.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -357,6 +357,11 @@ EthernetServer server(Port);
 #ifdef GatewayIP
 IPAddress gateway(GatewayIP);
 #endif
+#ifdef DNSIP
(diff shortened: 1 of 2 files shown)


fredlcore:
I just realized that now the default config file is no longer working because GatewayIP is no longer optional as it was before. Now people have to enter a gateway IP address which was not the case before.
The problem arises here:
```
#ifdef DNSIP
IPAddress dnsserver(DNSIP);
#else
IPAddress dnsserver(GatewayIP);
#endif

```
I have now added 
```
#ifdef GatewayIP
IPAddress dnsserver(GatewayIP);
#endif
and it compiles for me fine now and updated the repository accordingly. I hope it still works for you as well...
```

PR #43, opened 2019-12-03
Russian localization.

dukess:
1. First step to translate texts
2. create-undefs.pl can ignore comments in lang files
3. Moving units texts from BSB_lan_defs.h to lang files


fredlcore:
Thanks a lot, that seems like a lot of work!

PR #46, opened 2020-01-09
Russian translation + "external" web server

dukess:
I implemented web server for further use, ex. UI with AJAX. Need to test, because my arduino became buggy and it answers is wierd.


fredlcore:
Thank you for that, however, due to the memory constraints, I would like to keep additional functions to a minimum, even if they can be disabled through defines at the moment. 
Could you therefore re-submit your language file updates, so that I can accept and merge them separately? Thank you!

dukess:
FUI: I made proof of concept AJAX UI and export text strings to external files. This allowed to save from 24 (for German language) to 62 (current Russian translation) kB.

fredlcore:
Ah, ok, I see... But doesn't this require a major rewrite of the code? Maybe you can explain a bit further how you plan to do it.
As for the memory limitations, I'm currently working on porting the code to the Arduino Due which has double the flash capacity, so memory problems should not occur on that platform and it wouldn't require users to copy files to an SD card etc.

dukess:
I remember about your plans. :)
As I use JSON for data query UI not need to code rewrite but require to generate special language file. "Configuration", "Sensors", "Check new functions" just inserted into page and looks so:
изображение
My first goal is allow to user see list of functions (ready), values (partially ready), set new values (partially ready). 
Second goal: language switching on the fly.
Integration of service functions still for far future.

Ignore empty fields because it is problem with my arduino.
изображение

Show external resource. No problem.
изображение

fredlcore:
Ok, this looks really cool! Once you have it running stable, I'll consider adding it...

fredlcore:
After looking at your other repo more closely, I would really like to incorporate this feature, however if possible with minimal impact on a "standard" installation. As for functionality, I wonder whether the day of week display for the webserver is necessary, or if even having a directory index is necessary because people would probably not use this as an externally browsable web server (simply because it is so slow). So if people know which files are on the SD card beforehand, a directory listing could be dropped.

As for patching the script, maybe most of this could be dropped if it is properly included in the main code. The generation of the LANG_C could be done routinely in the create_undefs script which I run routinely with every change in the language files.

The ideal would be that people would only have to copy over a few files to the SD card and activate one or two definements. However, since I plan to move all configurations to a web-based configuration soon, it would be good if this would work also in such a context, as on the Due simply all definements will be activated but should still be possible to disable via the webinterface (I hope you understand my writing at this time of the day ;) ). Let me know how you would like to proceed.

dukess:
Hi! Sorry for long answer.
Day of week is part of "Last-Modified" header string. "Last-Modified" and "Content-Length" are two header with "cheap" calculations that will help the browser understand that the file on the web server has not changed, and it does not need to be downloaded again. In addition, the dayofweek function is already present in sketch. 
At this point in time I don't plan to add directory index yet because I do not need it. :)
But this version can send index.html instead internal web server when user request "/" URL. Thus, the user can choose the behavior of the device: if he prefers to work with the external interface, it is enough rename ajax.html to index.html. The home page of the internal server becomes unavailable. All other pages are still available. If the user prefers to interact with the internal server, then nothing need to do and the external interface is called by the name ajax.html.
I don't mind merging langtoajax.pl with create-undefs.pl. This decision will make support easier.
>The ideal would be that people would only have to copy over a few files to the SD card and activate one or two definements...
I will strive for this. :). Web-based configuration sounds cool, so i will to try to integrate with it.
My immediate plans are to test the latest version of the program and debug my patches for it. The fastjson patch speeds up the category generation process by 3 times and reduces the overhead, since println (F ()) sends only one byte per packet, unlike println (char *). It was not possible to achieve a significant acceleration of get values: we are limited by the bus speed.
The webserver patch is almost ready, but it has a strange behavior on my board: for some reason, one file has the wrong size. Most likely, this is an error on the SD card.
I have a request: it would be good to rename the broadcast files according to ISO 639. Now there are 4 such files:
LANG_cz,  "čeština" to LANG_cs
LANG_dk, "Dansk"to LANG_da
LANG_se, "Svenska"to LANG_sv
LANG_si, "slovenščina"to LANG_sl

Opened for discussion.
with best regards.

fredlcore:
Sorry for the delay in response, as you have seen, I have included all your changes now, including the renaming of the language files.

fredlcore:
As for the generation of the LANG_C.h, just let me know how this needs to be handled and I will include this in the automatic procedure when pushing to GitHub.

PR #55, opened 2020-01-28
Save some bytes.

dukess:
Add F() to some text strings.


fredlcore:
Thanks a lot, very helpful!

PR #57, opened 2020-02-08
Update BSB_lan.ino

dukess:
I found the cause of the problems going on with me. Everything turned out to be trivial: the size of the dev_unit variable is not enough to store language-independent unit identifiers.


fredlcore:
Thanks!

PR #58, opened 2020-02-10
Fast JSON

dukess:
Buffered output for JSON generation routine added. Category list built 3-4 times faster. Less network overhead.
Handler for VT_ONOFF and VT_YESNO type params added.


dukess:
Waiting for new version

PR #59, opened 2020-02-11
Update BSB_lan_defs.h

dukess:
Changed type from VT_BIT to VT_ENUM because ENUM5920, ENUM5922...ENUM5928 are equal.
изображение


fredlcore:
Thanks!

PR #63, opened 2020-02-17
External webserver, fast JSON generation

dukess:
This patch add these functions:
External web server for static content handling. Allow content caching, gzipping.
Fast JSON generation with buffered output. /JQ is little bit faster because here is protocol limitation (~300 ms per request).
New URL flag: "/W". Can be used with C and Q commands: /WC, /WQ. Generate web pages without header and footer. Useful for external interfaces.
Extend CAT_FEHLER up to func no. 6996 because JSON can send only 5 errors instead 20. 
Add two function for buffered output (may need to rewrite for wide-project usage): bufferedprint and bufferedprintln.


fredlcore:
Thanks a lot again, just went through all your code and this is really a great improvement!
Unfortunately, compilation fails on the Due with this error message:
```
/Users/frederik/.platformio/packages/framework-arduinosam/cores/sam/avr/pgmspace.h:86:50: error: cannot convert 'const char (*)[775]' to 'const char*' for argument '2' to 'char* strcpy(char*, const char*)'
 #define strcpy_PF(dest, src) strcpy((dest), (src))
```
I had to deal with this a while ago when doing the other transformations for the Due but maybe you have an idea at hand how to fix this? The underlying issue is that the Due does not know about "far" memory and just uses the normal strcpy function and therefore expects different parameters...

fredlcore:
Ok, this here did the trick:
```
#if defined(__SAM3X8E__)
                  strcpy(outBufp, ENUM_CAT+x);
#else
                  strcpy_PF(outBufp, pgm_get_far_address(ENUM_CAT)+x);
#endif
                  uint16_t y = strlen(outBufp);
                  x += y;
                  outBufp += y;
#if defined(__SAM3X8E__)
                  cat_min = ENUM_CAT_NR[cat*2];
                  cat_max = ENUM_CAT_NR[cat*2+1];
#else
                  cat_min = pgm_read_word_far(pgm_get_far_address(ENUM_CAT_NR) + (cat*2) * sizeof(ENUM_CAT_NR[0]));
                  cat_max = pgm_read_word_far(pgm_get_far_address(ENUM_CAT_NR) + (cat*2+1) * sizeof(ENUM_CAT_NR[0]));
#endif
```

dukess:
thank you very much for help!

PR #64, opened 2020-02-22
LANG_C.h generator

dukess:
what is to do better: add JS files builder to script or leave it in my project?


fredlcore:
No problem adding it to my create_undefs script. Is this the final version I can merge or do you still need to work on it?

fredlcore:
Also, what is the meaning of this:
```
# Save ~3 kB. Do not to change without discussion, please, because "_TEXT" used as marker in external interface.
```
What is it that should not be changed? And by whom?

dukess:
I use "_TEXT" substring for searching language resources. If we changed it to "_TXT" as example without changing AJAX interface then text will be displayed like this:
MENU_TXT_CFG<BR><BR>
MENU_TXT_VER: 0.43.38-20200131010227<BR>
MENU_TXT_RAM: 3314 Bytes <BR>
MENU_TXT_UPT: 346204569<BR>
MENU_TXT_BUS: BSB (66, 0) MENU_TXT_BRW<BR>
MENU_TXT_MMD: 0
<BR>MENU_TXT_VBL: 1<BR>
MENU_TXT_EXP: 
0 1 4 10 11 12 13 18 19 20 21 22 23 50 51 52 53 62 63 64 65 66 67 68 69 <BR>
MENU_TXT_MAC: 
00:80:41:19:69:90<BR>
MENU_TXT_AVT: <BR>
8700 - STR8700_TXT<BR>
8326 - STR8326_TXT<BR>
<BR>
MENU_TXT_LGP 
3600 MENU_TXT_SEC: <BR>
8700 - STR8700_TXT<BR>
8743 - STR8743_TXT<BR>
8314 - STR8314_TXT<BR>
<BR>

dukess:
> No problem adding it to my create_undefs script. Is this the final version I can merge or do you still need to work on it?

You can merge it. JS will be created in www subdir (creating automatically).

fredlcore:
But why should anybody change it to TXT? And why would it save 3 kB?

dukess:
> But why should anybody change it to TXT?
No.  No need to change now. But it is possibility to save flash memory in future.
> And why would it save 3 kB?
Because string (as example) "STR8314_TXT" (size is 12 bytes with zero) is shorter than "STR8314_TEXT" (13 bytes with zero plus 3 bytes alignment up to long type size). So lot of strings give us possibility to save memory.
Why 3kB? I just tried to compile sketch with  "_TXT" and "_TEXT". Flash image with "_TXT" was smaller.

fredlcore:
Ah, ok, get it. Then I would suggest we leave it with _TEXT for now.

fredlcore:
I think the 
```
mkdir "www";
```
is only necessary to run once, so I can remove it once the directory has been created here as it will be included in the GitHub repository?

dukess:
Yes, you can remove or comment it.
Thank you.

PR #65, opened 2020-02-25
www/languages.js autogeneration

dukess:
Generate list of languages into www/languages.js. For external interfaces


fredlcore:
Thanks!

fredlcore:
I have one small problem now: Whenever I run create_undefs.pl, the .gz files appear as "changed" files to be added to GitHub. However, without any changes to the actual language files, there shouldn't be any changes and gzip should (in theory) produce the same compressed file again, so that no change in the content would be detected. Do you know why this is the case and is there a way to fix this? Otherwise I'll have to deal with these changes everytime I routinely run create_undefs.pl.

dukess:
The first thing I can suggest is to generate a file with the names and dates of changes to the language files (LANG_*.h). And every time you run the script will check for changes to the date. If the file date has not changed, the script skips the file.

fredlcore:
Found it, the -N command line of gzip option was the problem.

fredlcore:
It stores the filename and timestamp data in the gzip header. Hope deflating the gzip still works, but I couldn't think of why it shouldn't...

PR #66, opened 2020-02-25
Modif create-undefs.pl

dukess:
rebuilding changed files only.
use Digest::MD5::File;


fredlcore:
Is this really necessary after I found the solution with the gzip options? Even if the files are recreated without need, that's ok as long as git doesn't regard them as new which seems to be the case now. If it's not necessary, I'd prefer to keep the code small.

dukess:
I don't know. I didn't find any information on the web that the absence of headers in a compressed file will somehow negatively affect browsers. So, most likely, it is not necessary. But I will keep this option for myself just in case.

PR #69, opened 2020-02-28
Add New Parameters And Define Some Unknown Command Ids

hypetsch:
This PR adds new parameters and defines some command ids currently marked as unknown.
Additionally some trailing white spaces are removed.


fredlcore:
Great, thanks a lot! As there are quite a few new parameters with data type VT_UNKNOWN, could you do a log of these parameters together with the value displayed on your heating system? Then it would be possible to set the right data type for these parameters.

hypetsch:
Sure, will try to do it asap.

fredlcore:
Thanks!

PR #74, opened 2020-03-03
/JC function, improved MQTT disabling.

dukess:
1. New function for JSON: Get list of possible values (like for a category) from user-defined list of functions. 
Example: /JC=505,700,701,702,711,1600,1602
2. Improved MQTT disabling.
MQTT uses a lot of memory even when it disabled.


fredlcore:
Sorry, I got it mixed up a bit, JC is not like JK, but rather JQ (as I wrote in the last paragraph). I don't see a difference there as I can call
/JQ=710,750,6100,8009,8308,8310,8314,8323,8326,8331,8700,8703,8704,8740,8744,8750,8830,8950,9009
and get a proper JSON output.

The error correction for the category is of course necessary as well as the Russian translations.

dukess:
Difference between /JQ and /JC is better visible on parameters with ENUM type.
/JC can be useful for management software. /JC return list of possible values, /JQ - current value
Example: if i want to make simple management interface with 3 SELECT (HK1 mode, HK2 mode, TWW mode) then i must:
with /JK and /JQ:
* download 3 categories (11, 13, 15), 
* extract parameters 700, 1000,1600 from categories. I need do it because i need human-readable parameters.
* download current values.

with /JC and /JQ:
* download parameters 700, 1000,1600, 
* download current values.

/JC=700,1600
{
  "700": {
    "name": "STR700_TEXT",
    "possibleValues": [
      { "enumValue": "0", "desc": "ENUM700_00_TEXT" },
      { "enumValue": "1", "desc": "ENUM700_01_TEXT" },
      { "enumValue": "2", "desc": "ENUM648_01_TEXT" },
      { "enumValue": "3", "desc": "ENUM700_03_TEXT" }
    ],
    "dataType": 1
  },
  "1600": {
    "name": "STR1600_TEXT",
    "possibleValues": [
      { "enumValue": "0", "desc": "MENU_TEXT_OFF" },
      { "enumValue": "1", "desc": "MENU_TEXT_ON" },
      { "enumValue": "2", "desc": "ENUM1600_02_TEXT" }
    ],
    "dataType": 1
  }
}

/JQ=700,1600
{
  "700": {
    "name": "STR700_TEXT",
    "value": "1",
    "unit": "",
    "desc": "ENUM700_01_TEXT",
    "dataType": 1
  },
  "1600": {
    "name": "STR1600_TEXT",
    "value": "0",
    "unit": "",
    "desc": "MENU_TEXT_OFF",
    "dataType": 1
  }
}

fredlcore:
Ah, ok, got it - not sure what kind of management software would use this, but it doesn't hurt, so I'll include it. Could you write 2-3 sentences what this command does for our manual? Thanks!

hypetsch:
A short question: Does /JQ now really return the language keys like `STR1600_TEXT` instead of human readable texts?

Just asking because this would break backward compatibility of the API (and additionally my openHAB binding which uses those texts).

fredlcore:
No, it will return the language selected in BSB_lan_config.h. @dukess however has added a feature that the language files can be offloaded onto SD card and the language keys will then be replaced by the language strings via JavaScript/AJAX. For this purpose, LANG_C.h has been added, but again, this only comes into play when you select "C" in the langauge definement.

dukess:
>  Could you write 2-3 sentences what this command does for our manual? 

For URL section in manual:
**/JC=\<x\>,\<y\>,\<z\>    JSON: Query possible values for parameters \<x\>, \<y\> and \<z\> for ENUM type parameters.** 
The format of the returned data is the same as the command /JK=\<x\>.
Unlike the /JQ command, it does not return the current parameters values.

For URL page:
/JC=\<x\>,\<y\>,\<z\>    JSON: Query possible values for parameters \<x\>, \<y\> and \<z\>. The format of the returned data is the same as the command /JK=\<x\>

dukess:
@hypetsch here is first result: https://github.com/dukess/bsb_lan_ajax

liudger:
I see I need to update home-assistant bsblan integration big time! (or actually just update the python bsblan library and the home assistant integrations will be fine)

fredlcore:
Oh, do these changes break your home assistant integration in any way? If so, please let us know!

dukess:
There should be no problems, the old functions have retained their behavior

liudger:
No it won't break. But I was thinking to build something for scanning the device and only add parameters that return a value. With the new interface it's a bit easier I guess.

PR #78, opened 2020-03-31
Small improvements

dukess:
1. Auto restart function (when defined TIMETORESTART time in ms).
2. VT_ONOFF & VT_YESNO can be 1 or 0 only
3. Return of I & S commands without header and footer (with W flag)
4. Piece of russian translation


fredlcore:
Thanks a lot again - a few small questions:

- Why the need for auto restart? That should normally not be necessary, and if so in special cases, such a function could/should be implemented in BSB_lan_custom.h which provides a timer for things like that.

- VT_ONOFF and VT_YESNO work when set to 0 and 1 but various heaters use 0 and 255 when sending this information. Also, looking at the commits, I don't see where this saves code, it looks like there are now more lines of code necessary (and thus more memory consumption)?

The rest looks fine for me.

dukess:
Hi!
* Sometimes my device freezes for an unknown reason after about 5 days of operation. The only reason i can see now: the hang-up occurs after using debugging via telnet. 
* I have a counter question: is it possible to find out exactly what value the hardware uses when returning: 1 or 255? For example, check a parameter or function when connecting? In my case (Baxi Luna Platinum), part of the parameters returns 1, and the other part returns 255. However, if I send back "1" in all cases, the device applies the parameters correctly.

dukess:
I created new data type just for YESNO and ONOFF. Unfortunally it breaks compatibility with old version but allows to split reliably behavior of ENUM and SWITCH.
I close request for testing purposes.

fredlcore:
As for the freezes: Yes, sometimes people have that, it seems to have to do with the Ethernet Shield, there is some discussion on the FHEM forum about it. The original shields seem not to be affected. In any case, this could and should be implemented via BSB_lan_custom.h and is therefore not going to be part of the main script.

As for YES/NO: It _seems_ to work fine when you send back a "1" even though the heater sends a 255, but I'm not sure if this really works across all machines. As for detecting it via hardware, I don't know how this should work. You could query a parameter and see if it is 1 or 255, but you would have to know beforehand which one is 1 or 255. Otherwise you would have to set a parameter temporarily to 1 and this may lead to unintended results.

As for transferring ON/OFF and YES/NO to a separate data type: What benefit would this bring? Compatibility should be maintained whereever possible.

dukess:
> As for the freezes: Yes, sometimes people have that, it seems to have to do with the Ethernet Shield, there is some discussion on the FHEM forum about it. The original shields seem not to be affected. In any case, this could and should be implemented via BSB_lan_custom.h and is therefore not going to be part of the main script.
Ok. 
For greater convenience, I suggest putting the reboot code in a separate function. If anyone needs it, they'll call it from BSB_lan_custom.
> 
> As for YES/NO: It _seems_ to work fine when you send back a "1" even though the heater sends a 255, but I'm not sure if this really works across all machines. As for detecting it via hardware, I don't know how this should work. You could query a parameter and see if it is 1 or 255, but you would have to know beforehand which one is 1 or 255. Otherwise you would have to set a parameter temporarily to 1 and this may lead to unintended results.
Yes, it's problem.
What I can suggest for external interfaces: if the parameter value is not zero, then send two telegrams, with 1 and 255. This is slower, but will probably work on all devices. BUT:
> 
> As for transferring ON/OFF and YES/NO to a separate data type: What benefit would this bring? Compatibility should be maintained whereever possible.
...for JSON it is possible only if we know that this parameter is a switch, not an enumeration. Otherwise, developers will have to write code for a handler that can still handle some situations incorrectly.
Idea! For JSON, we can add a field with the "isswitch" flag that is set if the data type YESNO or ONOFF.  This way we won't need to add a new data type and break compatibility.

fredlcore:
Yes, putting the reboot code into a seperate function is definitively a good idea.
As for YES/NO and ON/OFF, I was thinking for some time already to remove this special data type and convert it into an ENUM as well. Haven't thought this throgh 100% yet if this works, but would this solve the "switch" problem in JSON? Or would it make matters worse and still require the "isswitch" flag?

dukess:
> As for YES/NO and ON/OFF, I was thinking for some time already to remove this special data type and convert it into an ENUM as well. 

This way will only work if the devices return only 0 and 1. Since we can't be sure, it's best to leave it as it is.

>Haven't thought this throgh 100% yet if this works, but would this solve the "switch" problem in JSON? Or would it make matters worse and still require the "isswitch" flag?

Making it worse. I set "isswitch" flag always when i see data type YESNO or ONOFF. So if i can't recognize YESNO or ONOFF i can't set "isswitch" flag.

Can you review propose  https://github.com/fredlcore/bsb_lan/pull/80 ?

fredlcore:
Understood, have reviewed #80 and it looks fine for me. Just seem not to be able to merge it yet because of "work in progress".

And thanks, we're all still healthy here (in the countryside, there are not so many infections). Hope you are doing fine, too!

dukess:
thank you!
still healty :)
ready for merging.

fredlcore:
Glad to hear!
Just merged your commits but realized that at some point the "dump data payload for unknown types" has been commented out (3775++). Did you do this at some stage or was it someone else?

fredlcore:
Ok, got it, it's now handled in 2500++, sorry for the confusion!

dukess:
:)

PR #80, opened 2020-04-16
"error", "readonly" and "isswitch" variables

dukess:
Added "error", "readonly" and "isswitch" variables for enriching JSON.
Saving some RAM.


fredlcore:
Looks fine for me - although I cannot merge because it says "pull request is still a work in progress"?

BSB_lan.ino
```diff
@@ -6088,6 +6174,7 @@ uint8_t pps_offset = 0;
                   }
                   //client.println();
                   strcpy_P(jsonbuffer, PSTR("\n    ],\n"));
+                  if(decodedTelegram.isswitch == 1) strcat_P(jsonbuffer, PSTR("    \"isswitch\": 1,\n"));
```
hypetsch, line 6177: Just asking me: wouldn't it be a nicer API design to always add the `isswitch` member and instead set it to 0 or 1?

dukess: isswitch = 0 only for ENUMS or for any dataType?

P.S. I would prefer YESNO and ONOFF to be a separate data type (this would look nicer), but then compatibility would be broken. We discussed this with Frederik and traces of experiments can be found in this patch.

hypetsch: For all dataTypes.

Regarding the separate dataTypes: I was thinking of the same when implementing an openHab binding :)

dukess: I think it will be redundant, because for all other types it will be always 0.

Maybe in the future we can implement this after some big release.

fredlcore:
Thanks a lot for all the changes, merged now.

PR #84, opened 2020-05-05
New parameter 8004

chamtec:
Hello,

I add the 8004 ENUM data for cooling system.

Compilation OK
Test OK

Fred


fredlcore:
Thanks, but what is _11d_ supposed to be? Usually there is a match between the text marker and the value returned from the heater.

chamtec:
11d is for 285: Cooling mode Reduced according to RVS21 manual

fredlcore:
Ah, ok, got it, thanks!

PR #88, opened 2020-05-13
fix unresponsive website when terminating telnet session

loetmeister:
I had some issue on Windows (putty) using DebugTelnet, when disconnecting causing BSB-LAN to freeze (website was not loading at all, only ping was working)


dukess:
It works and helped to me

fredlcore:
Thanks!

loetmeister:
Thanks for the feedback.. and the merge. ;-) :1st_place_medal:

PR #89, opened 2020-05-14
Code optimization. Translation

dukess:
Duplicate strings ENUM8004_18_TEXT and ENUM8004_96_TEXT in language files and defs file . It is correct?

1. Code optimization. Saving 1900 bytes
2. Piece of Russian translation


fredlcore:
Thanks a lot! And yes, the doublettes must have resulted from the automatic conversion back then, thanks for spotting this!

PR #92, opened 2020-05-22
/JI command

dukess:
/J[SON]I[nformation] command. Show device configuration like /C URL.


fredlcore:
Thanks! What are the external interface texts? The ones in German still need some polishing (which I can do of course), but it would help to know what these are for?

PR #93, opened 2020-05-22
JSON broken. lost date and uptime values in datalog.

dukess:
Debug: 
1. JSON broken when user set empty lists. https://github.com/fredlcore/bsb_lan/issues/91#issuecomment-632908688
2. lost date and uptime values in datalog.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -427,7 +427,7 @@ char buffer[BUFLEN] = { 0 };
 /* buffer to print output lines*/
 #define OUTBUF_LEN  300
 char outBuf[OUTBUF_LEN] = { 0 };
-byte outBufLen=0;
+int outBufLen=0;
 
 char div_unit[32];
 
@@ -5861,7 +5861,7 @@ uint8_t pps_offset = 0;
 
           if (p[2] == 'I'){ // dump configuration in JSON
             free(jsonbuffer);
-            strcpy_P(formatbuf, PSTR("  \"version\": \"" BSB_VERSION "\",\n  \"freeram\": %d,\n  \"uptime\": %lu,\n  \"MAC\": \"%02hX:%02hX:%02hX:%02hX:%02hX:%02hX\",\n"));
+            strcpy_P(formatbuf, PSTR("  \"name\": \"BSB-LAN\",\n  \"version\": \"" BSB_VERSION "\",\n  \"freeram\": %d,\n  \"uptime\": %lu,\n  \"MAC\": \"%02hX:%02hX:%02hX:%02hX:%02hX:%02hX\",\n"));
             sprintf(outBuf, formatbuf, freeRam(), millis(), mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
             client.print(outBuf);
 // Bus info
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks!

dukess:
Thank you for the corrections in German and English!

fredlcore:
Hm, with the current build, I still get the JSON parsing error when having an empty log list?

dukess:
Can you drop here broken JSON?
It can look like this:
{
  "freespace": 474234,
  "loginterval": 3600,
  "logged": [
  ]
}

fredlcore:
It's again as it was before:
```
  "freespace": 1955054,
  "loginterval": 3600,
  "logged": 
  ]
}
```

fredlcore:
Somehow this line is missing from the original fix:
`if(!somethingexist) outBufLen = tempoutBufLen;`

fredlcore:
tempoutBufLen is now also missing, and outBufLen is now increased even if somethingexist is false:
```
            somethingexist = false;
            strcpy_P(formatbuf, PSTR(",\n  \"freespace\": %lu,\n  \"loginterval\": %d,\n  \"logged\": [\n"));
            outBufLen+=sprintf(outBuf+outBufLen, formatbuf, SD.vol()->freeClusterCount(), log_interval);
```
The same may be the case if avg_parameters are empty.

dukess:
Add 
Stupid error: always cutted last two symbols.
            if (numLogValues > 0) outBufLen-=2; //two bytes shift (delete comma and \n)
and
            if (numAverages > 0) outBufLen-=2; //two bytes shift (delete comma and \n)
instead 
            outBufLen-=2; //two bytes shift (delete comma and \n)

fredlcore:
This still won't work because numLogValues is by default 20, and so is numAverages. It should probably be log_parameters[0]>0 instead of numLogValues>0, shouldn't it?

dukess:
you right.

dukess:
with ipwe_parameters[] it will be work, but not with  log_parameters

fredlcore:
Is ipwe_parameters dumped in /JI?

dukess:
Not now, but it's easy to add

fredlcore:
No need, I guess, I was just wondering because you mentioned that it would work with the above.

dukess:
I mean ipwe_parameters has not hadrcoded size, so sizeof(ipwe_parameters) / sizeof(int) can be used for actual size calculation.
By the way, why log_parameters and avg_parameters have fixed size? It is legacy or for some reason?

fredlcore:
Actually, I probably just forgot that I did it that way with ipwe_parameters ;). We could change it to this way with the other parameters as well, but since one of my goals is to have the whole of BSB-LAN entirely configurable via the web-interface so that we can distribute pre-compiled binaries, we will have to allocate a fixed number of entries anyway so that it can be stored in the EEPROM.

dukess:
Understand! 
P.S. We can make EEPROM dynamic allocation, it could save space, but addressing will be more complexity.

fredlcore:
I think it would be fine if we give people a maximum number of entries for each array. On the other hand, the EEPROM could be divided into two parts: One that has fixed allocations for configuration variables that are only set/read by us programmers and one part that stores the arrays such as log_paramters. This part could again be divided into two segments where the first segment stores the length of each array used in the configuration webinterface, and the second segment the actual data, and when configuration is written, both segments will be updated accordingly and can be read accordingly later on. Just an idea...

PR #97, opened 2020-05-25
buffer IPWE output + small clean-up

loetmeister:
Can someone please validate the IPWE output, too? :-)
I tested with OneWire, but no DHT Sensor.
DHT Sensor output is now in one line, using column Temperature and Humidity.


fredlcore:
Thanks, but this will break parsing for DHT sensors because humidity sensors in the original IPWE device had a "F" for "Sensortyp" instead of a "T" for temperature sensors, and I'm not sure if people or external modules are matching against this, although I see that your approach makes more sense given that there is a temperature as well as a humidity row.

Also one more question: How much space in terms of flash or RAM does this bufferedprint/sprintf approach actually save? Since the move to the Due, I'm less inclined to save 3 bytes here and 5 bytes there if code readability is significantly decreased. (And yes, I know most of my code is already hard to read due to lack of comments which is why I generally value any contributions despite this fact ;).) Debugging something where the format string is in one place and the population with variables in another and the actual output in a function even further apart, does not make things really easier IMHO...

loetmeister:
Thanks.... yes parsing would break... even if I change it back to "F"? - I still would keep "temperatue and humidity in a single line. :-|

The change was not so much about code reduction, but use of `bufferedprint()` or `outBuf` to speedup the HTLM rendering. However it may saved 100 odd bytes...

The code would be nice to read, with one HTLM line + sprintf().... but the Arduino lib does not allow type float variables in sprintf(), so I had to break it into some pieces, unfortunately. Maybe there should be some warpper function to allow for signed float values (like `_printFIXPOINT()`)?

fredlcore:
Thanks for the explanations; as for the speed improvement, is there any indication as to how much this improves? Like double as fast or just 5% faster? Given that the main bottleneck is still the query of each parameter over the bus (except for external sensors), I doubt that the improvement will be noticeable, wouldn't it?

dukess:
I think there is no point in replacing the "old" with the "new", since the string length is only one byte.
Old:   if (tx_msg[i] < 16) client.print(F("0"));
New:  if (tx_msg[i] < 16) bufferedprint(client, PSTR("0"));

By the way i deleted "client" variable from bufferedprint and bufferedprintln because "client" is global.

loetmeister:
> I think there is no point in replacing the "old" with the "new", since the string length is only one byte.
> Old: if (tx_msg[i] < 16) client.print(F("0"));
> New: if (tx_msg[i] < 16) bufferedprint(client, PSTR("0"));

I expected no difference, too. However it saves 4 bytes for each function call. Also I replaced all client.print(F("xxxx")); with bufferedprint(client, PSTR("xxxx")); in the entire sketch, to have it consistent.

> By the way i deleted "client" variable from bufferedprint and bufferedprintln because "client" is global.

yes, looks more clean. I guess the compiler would have it optimized anyway. :-)

loetmeister:
> Thanks for the explanations; as for the speed improvement, is there any indication as to how much this improves? Like double as fast or just 5% faster? Given that the main bottleneck is still the query of each parameter over the bus (except for external sensors), I doubt that the improvement will be noticeable, wouldn't it?

I did a quick test, by counting all packets send during a full ipwe.cgi site refresh. I just reverted the code for DHT22. This gave a 155% increase. (approx. 240 vs 610 packets - HTTP & TCP)
I think it would be even less, as I generated 4 rows for DHT sensor, instead of 2... quick and dirty code, but to have at least simlar HTLM code :-)

Old code:
```
     client.print(F("<tr><td>T<br></td><td>"));
     client.print(counter);
     client.print(F("<br></td><td>"));
     client.print(F("DHT sensor "));
     client.print(i+1);
     client.print(F(" temperature"));
     client.print(F("<br></td><td>"));
     client.print(temp);
     client.print(F("<br></td><td>0<br></td><td>0<br></td><td>0<br></td></tr>"));
     counter++;
     client.print(F("<tr><td>F<br></td><td>"));
     client.print(counter);
     client.print(F("<br></td><td>"));
     client.print(F("DHT sensor "));
     client.print(i+1);
     client.print(F(" humidity"));
     client.print(F("<br></td><td>0<br></td><td>"));
     client.print(hum);
     client.print(F("<br></td><td>0<br></td><td>0<br></td></tr>"));
```

vs
```
      outBufclear();
      outBufLen+=sprintf(outBuf, formatbuf, type, counter, i+1);
      _printFIXPOINT(temp,2);
      strcat_P(outBuf, PSTR("<br></td><td>0<br></td><td>0<br></td><td>0<br></td></tr>"));
      client.println(outBuf);
       counter++;
      memcpy_P(type, PSTR("F"), sizeof(type));
      outBufclear();
      outBufLen+=sprintf(outBuf, formatbuf, type, counter, i+1);
      strcat_P(outBuf, PSTR("0<br></td><td>"));
      _printFIXPOINT(hum,0);
      strcat_P(outBuf, PSTR("<br></td><td>0<br></td><td>0<br></td></tr>"));
      client.println(outBuf);
```

dukess:
> However it saves 4 bytes for each function call.

Ok. 

> Also I replaced all client.print(F("xxxx")); with bufferedprint(client, PSTR("xxxx")); in the entire sketch, to have it consistent.

But be careful: buffer variable is used in query() and some other function so here can be intersections.

loetmeister:
> But be careful: buffer variable is used in query() and some other function so here can be intersections.
> 
> This is the reason why I didn't do buffered output in the entire program. Maybe I'm being too careful.

Yes... true. There seem to be the old style, where `buffer` and `outBuf` is used, to concatenate the output and then send it to the network client (Ethernet Shield) in bigger chunks.
Or the "hacky" style, to print it to the network client as it comes.... (which is ok, as you get quickly get lost what wrote data to a global buffer variable - at least this is true for me)

You can try what I broke, using `bufferedprint` extensively... :-)
https://github.com/loetmeister/bsb_lan/tree/mega2560

fredlcore:
When I first dealt with the code, I was also worried how to keep track of all the different functions working on the buffer and outBuf variables, and in order not to break anything that will in the end be hard to find, I chose to go the "client.print()" way where possible. 
Thanks for speed tests, it certainly would make sense to go that route, but only if we can make sure that there are no interferences between the different functions. To be honest, I don't have the time to check and/or test this at the moment...

fredlcore:
What happens to:
`          client.println(F(MENU_TEXT_AVT ": <BR>"));`

loetmeister:
> What happens to:
> ` client.println(F(MENU_TEXT_AVT ": <BR>"));`

I removed too much... not intentionally :-)

I also changed DHT output to old behaviour: Two rows per Sensor.
https://github.com/fredlcore/bsb_lan/pull/97/files#diff-1118998d47d0c636c492e49a080339afL4243-R4297

dukess:
I believe that for deep optimization it is necessary to separate the decoding of telegrams and their printing. In the "decoding" part, we can directly decode the telegram, output debugging information, and save data to the structure: the resulting value in a string variable (for a timer , three variables), units of measurement, variable type, error status, and other flags. In the "print" part, we can build a string to output from the structure data and print it directly.
HTML output may be somewhat more cumbersome, but JSON and MQTT output will be easier. We may be able to remove one of the buffers or use it as an output buffer without fear of damaging the data.

But it will be a revolution.

fredlcore:
A revolution indeed. Not that I would be against it, but I rather follow the directive "never change a running system", especially when the benefits are rather unnoticeable. In my opinion, there are more important things on the ToDo list, such as a complete web-based configuration that makes the _config.h at least completely optional. Or a web-based firmware-update (not sure if that would work with the Due). Or...

dukess:
In fact, the main advantage will be the acceleration of the web interface when using buffering.

>Or a web-based firmware-update (not sure if that would work with the Due).

Looks cool. Has anyone done this for any other project?

loetmeister:
> In fact, the main advantage will be the acceleration of the web interface when using buffering.

Yes, not much visible or immediate improvement, I would agree. Potentially easier code maintenance or more flexible/robust code for next featues...

fredlcore:
> > Or a web-based firmware-update (not sure if that would work with the Due).
> Looks cool. Has anyone done this for any other project?

Apparently it's not as easy as with the ESP or even Arduino Uno/Mega:
https://forum.arduino.cc/index.php?topic=593800.msg4035803#msg4035803

dukess:
Thank you!

dukess:
Comrades, I decided to stretch my brain and redo the program as I described above. A little testing showed that the idea works, but there may be errors in the code.
https://github.com/dukess/bsb_lan/commit/8344657b1f0a3eaed116f74cee7ac1c88229a378

loetmeister:
Wow, that's a bigger change... I did not look at all the details, but it seems to give more structure.
I think _printfixpoint can now be used easier for ipwe, as it's done for the other output of DHT an one-wire sensors.

Would it make sense to have an outBuf print function, that clears the buffer after printing? Or a function to add data to the buffer, it can handle the buffer length variable and print the buffer, if the new string would not fit anymore?

fredlcore:
Thanks, @dukess, as always: great work! I only manage to grasp the larger context of your approach, but don't have the time to check every change you made, but this way would break compatibility with previous usage of the query() function that people might have used in the _custom.h files when they query parameters for custom processing. 

Having a struct is definitively an advantage from a programmatic perspective, but the benefits would be rather for a small group of people (i.e. us two or three people working on the code ;-) ), and all others who use the query() function in the previous fashion would encounter compiler errors in the update process.

dukess:
> Would it make sense to have an outBuf print function, that clears the buffer after printing? Or a function to add data to the buffer, it can handle the buffer length variable and print the buffer, if the new string would not fit anymore?

I think it could be next step.

dukess:
>  but this way would break compatibility with previous usage of the query() function that people might have used in the _custom.h files when they query parameters for custom processing.

Indeed, I forgot about backward compatibility. However, I can try overloading the function and returning the pointer to the buffer.

fredlcore:
It's not just what the function returns, but also how it is called. 
As I said before, I, personally, would not invest a lot of time and efforts to fix or improve code that is working and where people know how it is used in their functions - "never change a running system".
If it is something that you two enjoy working on, then please, go ahead, but to me there are more important parts of the project where the energy invested would have more impact than saving a few hundred bytes...

dukess:
When I started this job, I had two goals: to train my brain and to understand the program in more detail. I can say that I achieved both goals. :-) Some code from the patch can be used in the main branch without radical changes. But the main thing for me is that I now know how to add output of measurement units to JSON at the stage of category output.

Evolution is our way. :)

fredlcore:
Ok, that's a great motivation :) - and again, I really like the new approach, it is much cleaner and probably allows for easier implementation of other features. But I just had another case of someone who migrated from the Mega to the Due and the code compiled fine, but it wasn't working because he didn't change the bus variable from 68/69 to 19/18. That took quite some time to figure out ;). So I guess when we change the query function is changed in the way it is called, people will just tell me "it's not working/compiling anymore!", even when we put it prominently on the website.

So for maintaining compatibility - where somehow possible - would be in my very own interest ;). We can, however, keep these changes in the back of our minds (or better: a list) and once we have a change where we have to break compatibility, we can make all these changes at the same time...

dukess:
Yes, I understand your concerns. Debugging non-obvious cases is a torment. I made another patch that doesn't break the compatibility of functions and variables. I added something new, but it works in parallel with the old one, without interfering.
https://github.com/dukess/bsb_lan/commit/c727c89c56aa4afa9dcd7e0246468a2d70482df1

loetmeister:
Thanks guys. I guess I will close this pull request. Should I open a new issue (enhancemen) on this topic, to not loose the conversation?

fredlcore:
I think closing this for the time being makes sense. Enhancement of course are always welcome, but the above should be taken into account.

loetmeister:
opened #102

PR #100, opened 2020-06-01
Expanding /C output, JSON improvement, memory optimization.

dukess:
Print list of installed modules in /C command
decodedTelegram structure was expanded. It used for progs parameters and value storing.
_printFIXPOINT() function added for using with buffers other than outBuf.
Reading Prog parameters was moved to separate function.
Some memory optimization: char * to PSTR.
JSON: "unit" field sent with /JK, /JC commands now.  Typo: numbers values sending without quotation now. Bug removed: "readonly" was sending twice.
SdFat 1.1.4, checking free space on SD before logging
Print list of installed modules in /C command


fredlcore:
Thanks, could you do me a favor and test (in general) if these changes compile for the Due as well without any errors or warnings? That would save me quite some time especially with larger diffs.

dukess:
Tried to compile for Due. The compiler returned an error "'class SDLib::File' has no member named 'dirEntry'".  dirEntry is used to get file properties in WEBSERVER.  
When I replaced SD.h with SDfat.h the project was compiled successfully. (Of course, if user disable WEBSERVER, the project will be compiled with SD.h)
Are there any reasons to use SD.h?

```
#if defined(__SAM3X8E__)
//  #include <SD.h>
  #include "src/SdFat/SdFat.h" // if you run into troubles with SdFat.h, just remove these two lines and activate the line above.
  SdFat SD;
#else
```

fredlcore:
Very strange. I was using SdFat for the Mega all this while and when I ported the code for the Due, it created problems (which I forgot which these were) which forced me to switch back to SD.h. But now it compiles and runs fine with SdFat.h (although I don't do SD logging at the moment). So you can change it so that all platforms use SdFat.h, and if then all compiles well, I'll merge this pull request.

dukess:
By the way, what you think about upgrading SdFat library? Latest version 1.1.4, diff log: https://github.com/greiman/SdFat/compare/1.0.1...1.1.4

I tried compile it for Mega and Due - and have no problem. But it need to change some pathes. 
For at least one night, the new version is running successfully on Mega.

fredlcore:
Yes, that would probably be wise, but you are right, a few paths have to be adjusted which is why I always prioritize this rather low as long as things are running fine ;)...

dukess:
I can add these changes to pull request if you wish :-)

fredlcore:
Oh, that would be awesome!

dukess:
Done!

fredlcore:
Hm, compilation on the due breaks with this error:
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:1649:62: error: cannot convert 'char*' to 'float' for argument '1' to 'void _printFIXPOINT(float, int)'

dukess:
Moment. Here is two functions now:
```
void _printFIXPOINT(float dval, int precision){
  outBufLen+=_printFIXPOINT(outBuf+outBufLen, dval, precision);
}
int _printFIXPOINT(char *p, float dval, int precision){
```
but with same text source no problem

fredlcore:
I see, but does it compile for you without errors? For me it fails for both the Due and the Mega, compiling under VS Code / PlatformIO...

dukess:
I use Arduino studio :-) so here can be problem.
Looks like your IDE can't support function overload.
The solution then will be this: I will rename the function I added to eliminate this type of problem.

dukess:
https://github.com/fredlcore/bsb_lan/pull/101

Now I understand that this would be a big problem with the query() function.

fredlcore:
I could live with it not compiling under VS Code, but if it also does not compile under the Arduino IDE, then we would have a big problem ;)...

dukess:
I will wait for news

fredlcore:
Have pushed my changes to GitHub just now...

dukess:
👍

fredlcore:
So now we can use the 'new' query() function wherever we want in the code and still have the legacy query() for users who already use it in their code.

dukess:
I think it can be done a little later. Let the code stabilize. :-)

fredlcore:
👍

fredlcore:
I just checked on my system with an SD card inserted, I get free size: 0 - the card works and logs data to is, so ther seems to be an error here?

fredlcore:
Correction, it does not log anymore and gives me an error setting up SD card on startup. And now I also remember why I had to use Sd.h instead of SdFat.h on the Due: It simply would not detect the SD card on startup. Although I thought I had tried it yesterday where it worked with SdFat.h as well, so maybe its the version upgrade?

fredlcore:
Ok, found the error - SD.begin(4) was hardcoded to the Mega's SPI pin. Omitting the '4' will now automatically select the correct bin based on the board (at least I hope so for the Mega). Works now with the Due at least in the most current Master version.

fredlcore:
Grr... It was working just now, but now it isn't and I can't tell for certain whether it was with the new SdFat library or the one from before. Will have to check tomorrow with a clearer mind...

dukess:
Sad situation. I try to find differences between libraries.
First look give me this:
trouble with SD on Due (Clock frequency too high):
https://forum.arduino.cc/index.php?topic=323454.0

/** Set SCK rate to F_CPU/3 for Due */
#define SPI_DIV3_SPEED SD_SCK_HZ(F_CPU/3)
/** Set SCK rate to F_CPU/6 for Due */
#define SPI_DIV6_SPEED SD_SCK_HZ(F_CPU/6)

You can try to set
SD.begin(SD_CS_PIN, SPI_DIV3_SPEED ))
or
SD.begin(SD_CS_PIN, SPI_DIV6_SPEED ))

fredlcore:
Wow, you are awesome, how come I did not find this?
What I'm wondering is this:
`SD.begin(SD_CS_PIN, SPI_DIV3_SPEED )); // full speed`
Why is DIV3 full speed? shouldn't it be 1/3 of speed?
SPI_HALF_SPEED as mentioned in that thread also works. 
I wonder now which one would be fastest, DIV3, DIV6 or HALF?

dukess:
In short, if i understand correctly, SD card depends on CPU clock. So, since Duo is faster, we need to add a divider so that the card can keep up with the clock. 
SPI_DIV3_SPEED  - full speed of SD card for Due. Maximum throughput.
SPI_DIV6_SPEED  - half speed of SD card for Due. Used when wiring between board and SD card is more than 10 centimeters (+-).

fredlcore:
Ok, but what I don't understand here in this part:
```
#define SD_SCK_HZ(maxSpeed) SPISettings(maxSpeed, MSBFIRST, SPI_MODE0)
#define SD_SCK_MHZ(maxMhz) SPISettings(1000000UL*maxMhz, MSBFIRST, SPI_MODE0)
// SPI divisor constants
/** Set SCK to max rate of F_CPU/2. */
#define SPI_FULL_SPEED SD_SCK_MHZ(50)
/** Set SCK rate to F_CPU/3 for Due */
#define SPI_DIV3_SPEED SD_SCK_HZ(F_CPU/3)
```
is why F_CPU/2 is "max rate" and how F_CPU/3 should be full speed - somehow that does not make sense to me, especially when F_CPU/3 is working for me now and the default value SPI_FULL_SPEED set in the SD.begin() function is not? So F_CPU/3 can't be full speed then?

dukess:
I think it just need to remember, otherwise, you need to read and understand a lot of sources and descriptions of hardware. :-)

fredlcore:
Ok, but I still don't understand how F_CPU/3 is full speed for Due when F_CPU/2 is actually faster?

dukess:
I don't know myself, but I can assume that with F_CPU/2 the bus operation may be unstable. Or when F_CPU/3, we can use DMA. My knowledge is not enough, so can only rely on the developer for this issue.

PR #101, opened 2020-06-04
Fast fix for _printFIXPOINT

dukess:
remove _printFIXPOINT function overloading


fredlcore:
Found the problem - PlatformIO does not like forward declarations. So all I had to do was swapping the two functions (and rename printFIXPOINTtoBuf back to printFIXPOINT and now it's working.

dukess:
My mistake :-(

PR #105, opened 2020-06-06
Fix spelling

liudger:
authentification comes from french it's somewhat correct but authentication is better.


fredlcore:
Thanks!

PR #108, opened 2020-06-11
improvements: time output in func Nr 500-616, old bug in category #32

dukess:
VT_ONOFF, VT_YESNO, VT_CLOSEDOPEN, VT_VOLTAGEONOFF got "isswitch" flag for JSON and enums instead strings -> code much cleaner
old bug (pre-0.44): Error messages (category #32) not shown correctly. ENUM_ERROR was assigned to every "empty" VT_ERRORCODE .
query() function was splitted as in "revolution" but with backward compatibility.
Web interface: time output in func Nr 500-616 was corrected.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -1648,7 +1648,10 @@ void __attribute__((deprecated)) printSINT(byte *msg,byte data_len, long multipl
      decodedTelegram.error = 1;
    }
  }
-
+ //deprecated
+ void __attribute__((deprecated)) printDWORD(byte *msg,byte data_len, long divider, const char *postfix){
+   printDWORD(msg, data_len, divider);
+ }
 /** *****************************************************************
  *  Function:
  *  Does:
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -525,12 +525,12 @@ int loopCount = 0;
 
 struct decodedTelegram_t {
 //Commented fields for future use
-//int cat; //category number
-//uint32_t prognr; //program number
-//uint_farptr_t cataddr; //category description string address
-//uint_farptr_t prognraddr; //prognr description string address
+int cat; //category number
+uint32_t prognr; //program number
+uint_farptr_t cataddr; //category description string address
+uint_farptr_t prognraddr; //prognr description string address
 uint_farptr_t enumdescaddr; //enum description string address
-uint_farptr_t enumstr; //first element of enum address
+uint_farptr_t enumstr; //address of first element of enum
 uint16_t enumstr_len;  //enum length
 uint8_t msg_type; //telegram type
 uint8_t tlg_addr; //telegram address
@@ -1183,10 +1183,10 @@ void loadPrognrElementsFromTable(int i){
   uint8_t type=get_cmdtbl_type(i);
   uint8_t flags=get_cmdtbl_flags(i);
   uint8_t div_type=1;
-  uint8_t div_precision=1;
+  decodedTelegram.precision=1;
   uint8_t div_data_type=0;
   uint8_t div_unit_len=0;
-  float div_operand=1;
+  decodedTelegram.operand=1;
 
   if ((flags & FL_RONLY) == FL_RONLY)
     decodedTelegram.readonly = 1;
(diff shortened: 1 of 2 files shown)


fredlcore:
Thanks! I hope it will work stable :)...

dukess:
Me too!

fredlcore:
Again, there are several errors when compiling after including this request, I think it's again the isue of forward declarations. I hope I can fix these...

fredlcore:
Ok, got them, but there are a few warnings with unused variables left, can you please check?
```
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino: In function 'void printTelegram(byte*, int)':
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:2610:23: warning: unused variable 'p' [-Wunused-variable]
                 char *p = outBuf+outBufLen;
                       ^
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:2655:28: warning: unused variable 'len' [-Wunused-variable]
                   uint16_t len=0;
                            ^~~
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino: In function 'void query(int)':
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:4019:9: warning: unused variable 'pvalstr' [-Wunused-variable]
   char *pvalstr = NULL;
```

dukess:
Can you drop log here?

fredlcore <notifications@github.com> 11 июня 2020 г. 23:43:30 написал:

fredlcore:
Sorry, the compiler window with the errors is closed already and the errors are now fixed after moving printCHOICE below printENUM. The warnings above still apply. I don't want to just delete these lines because maybe these variables were meant for something else?

dukess:
Difference in compilers as usually.
Better way is make header file with functions declarations.

>Ok, got them, but there are a few warnings with unused variables left, can you please check?

These variables can be deleted without any regret

fredlcore:
Ok, thanks, done :).
But by the way: create_undefs.pl throws this warning:
```
LANG_RU.h WARNING: string length MENU_TEXT_VB3 too big: 414 bytes.
```

dukess:
Thank you!
All strings using now "outBuf" instead "buffer" so
@size = grep(/#define BUFLEN /, @Ino);
$size[0] =~ /#define BUFLEN (.*)/;
can be replaced by 
@size = grep(/#define OUTBUF_LEN /, @Ino);
$size[0] =~ /#define OUTBUF_LEN (.*)/;

fredlcore:
Ok, thanks, done.

dukess:
ENUM6620 and ENUM6621 looks like a choice. Propose to change their type to VT_YESNO.
https://github.com/dukess/bsb_lan/commit/5629bf84a11c9f2d4a03d58cbf16395d6706df7c

fredlcore:
Hm, in my manual it is a choice between "Segment" and "System", so not really a Yes/No choice?

dukess:
It is not YES/NO, but behavior looks like a switch: tree possible states (0,1, 255) with two real states (zero and non-zero). I don't insist - you know protocol better.

fredlcore:
It is in the end a two state parameter, but when we switch it to YES/NO, it will display a "yes" and a "no" in the web display instead of "System" and "Segment". So it won't really make sense. But if in your new version you can change the displayed text from YES/NO to some other text, then of course you are right.

But one more thing I saw in that code: We know that in most (if not all) cases 1 and 255 are both "YES" (and probably everything in between that has bit 0 set to 1), but I think we still need to keep both 1 and 255 because we don't know which system sends a 1 and which sends a 255. So when we get a 255 in the data telegram and match it with the ENUMs, it will return undefined because it can't find a match. It could maybe be done by ANDing the received value with 0x01 and match it then against the ENUM.

dukess:
>It is in the end a two state parameter, but when we switch it to YES/NO, it will display a "yes" and a "no" in the web display instead of "System" and "Segment". So it won't really make sense. But if in your new version you can change the displayed text from YES/NO to some other text, then of course you are right.

Currently, the strings for VT_YES NO, VT_ON OFF, etc are stored as ENUM and written in the function table, so they will be get depending on the program number and no problem.

>But one more thing I saw in that code: We know that in most (if not all) cases 1 and 255 are both "YES" (and probably everything in between that has bit 0 set to 1), but I think we still need to keep both 1 and 255 because we don't know which system sends a 1 and which sends a 255. So when we get a 255 in the data telegram and match it with the ENUMs, it will return undefined because it can't find a match. It could maybe be done by ANDing the received value with 0x01 and match it then against the ENUM.

Yes, I remember this specificity, so new printCHOICE() function checks the value (zero or not zero) and returns the first or second row from the enumeration. 
Of course, the user will see the actual return value.

fredlcore:
Ok, great, then please go ahead!

dukess:
New pull was created 
 https://github.com/fredlcore/bsb_lan/pull/110

fredlcore:
Thanks, merged already.

PR #110, opened 2020-06-13
Intensity of  I/O operations was reduced. Changing type for ENUM6620 and ENUM6621

dukess:
printCHOICE() calling was rewritten.
averages.txt deleting and creating was changed to seek(0);
ENUM6620 and ENUM6621 looks like a choice. Propose to change their type to VT_YESNO.
Intensity of  I/O operations was reduced for average parameters: write file one time per minute instead write file every loop


fredlcore:
Thanks!

PR #111, opened 2020-06-19
Update BSB_lan.ino

dukess:
Bugfix: in case of error shown value from category ENUM.
Bugfix: typo error in determining the device type.
Bugfix: in case of error telegram dump not shown.
Error codes rewritten.
LogTelegram() write to separate file (journal.txt). It can be recreated with /D0 (same time with datalog.txt) command and dumped with /DJ command.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -431,6 +431,7 @@ int outBufLen=0;
 
 #define averagesFileName "averages.txt"
 #define datalogFileName "datalog.txt"
+#define journalFileName "journal.txt"
 
 #ifdef WIFI
 WiFiEspClient client;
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks, can't check it so quickly, but why the separation between journallog and datalog and what is the difference?

dukess:
Propose to use datalog.txt for selected parameters logging and chart drawing. journal.txt use for write contents of telegrams according to the settings.
By the way: I tried very hard to enable logging: i ran commands /LU=0, /LB=0, but it didn't work. Only after uncommenting parameter 30000 log started to fill with data.

fredlcore:
Hm, I didn't know that lines such as 
39026;19.06.2020 14:08:55;6;LAN->HEIZ QUR;DC C2 00 0B 06 3D 05 00 9D 20 DE
are being logged currently in datalog.txt. I thought these telegrams are only written to the (debug) console?
Generally logging all these telegrams will quickly fill up the SD card (on BSB at least one pair of telegrams every 10 seconds.

Logging unknown telegrams has to be enabled via the webinterface (/LU=1) and at least for me it was helpful to see these in the context of other telegrams going in and out.

To start logging via the webinterface, you only have to use parameter /L like in
/L=60,8300,8700
to start logging parameters 8300 and 8700 every 60 seconds.

dukess:
>Hm, I didn't know that lines such as
39026;19.06.2020 14:08:55;6;LAN->HEIZ QUR;DC C2 00 0B 06 3D 05 00 9D 20 DE
are being logged currently in datalog.txt. I thought these telegrams are only written to the (debug) console?

The contents of telegrams are written to a file only if the first value in the log_parameters array is 30000. This is specified in BSB_lan.config, but I didn't pay attention and only realized it after reading sketch code.
`//30000,                  // Logging von "rohen" Bus-Datentelegrammen (macht nur als alleiniger Parameter Sinn)`
I assume you have this option enabled. 
I don't know why logging is enabled this way, but it doesn't seem quite logical to me.

>Generally logging all these telegrams will quickly fill up the SD card (on BSB at least one pair of telegrams every 10 seconds.

After updating SdFat to 1.1.4 we have protection against card overflow.

>To start logging via the webinterface, you only have to use parameter /L like in
/L=60,8300,8700
to start logging parameters 8300 and 8700 every 60 seconds.

I use this function with great pleasure. :)

dukess:
If you still have doubts about keeping telegram log in a separate file, i can change #define journalFileName "journal.txt" to #define journalFileName "datalog.txt" to accept error correction as soon as possible. We can postpone this discussion for a while.

fredlcore:
You know BSB-LAN better than I do now ;)...
30000 was introduced to log bus telegrams in general. With /LU=1 you can limit logging to unknown telegrams which can be used to only find telegrams that still need to be decoded.

You are right, it makes more sense to separate the telegrams and the data. So let's do it :)...

dukess:
Thank you!
I will write Ulf a letter about what needs to be corrected in the documentation.

PR #114, opened 2020-06-28
Automatic creation of datalog.txt

dukess:
Automatic creation of datalog.txt on startup if it is not present on the SD card


fredlcore:
Thanks!

PR #115, opened 2020-06-29
reset for Due. NOT TESTED "in vivo"

dukess:


fredlcore:
Thanks, will try!

fredlcore:
Unfortunately, it does not work. Due hangs after performing this reset. I think I tried the solutions in this thread before and none of those seemed to have worked, but I don't know why. Maybe you can try with the real one once you have the Due?

fredlcore:
Oh, and I checked the thread you were linking to, and at the end they actually mention
`  rstc_start_software_reset(RSTC);`
and not
`  rstc_set_external_reset(RSTC, 256);`
as you submitted. I tried with the first one and now it works!
Any reason why you went for the second one?

dukess:
can you try rstc_start_software_reset(RSTC);
This is function from forum thread without restart ethernet shield.

fredlcore:
Yes, that's the one I meant, that one worked. Didn't know that the other one would reset the Ethernet shield. However, it's not the Ethernet shield that is the problem because also when I have the Due on my serial monitor, it just hangs there after the reset command, no prior-to-ethernet debug messages etc.
But as long as the other command works, I think we're fine...

dukess:
rstc_set_external_reset(RSTC, 256); must reset external devices too. It may need to select the duration value, because `RSTC->RSTC_MR = 0xA5000801; // Set RST pin to go low for 256 clock cycles on reset`
 0801 is not like 256

fredlcore:
Ah, ok, so what value should I try? Longer or shorter?

dukess:
Hmmm... 2048?

fredlcore:
Tried 2049 (as per 0x0801), no luck...
Does this actually perform a reset or just set the time for the reset pin?

fredlcore:
10 also doesn't work...

dukess:
Found one example: 
https://cpp.hotexamples.com/ru/examples/-/-/rstc_set_external_reset/cpp-rstc_set_external_reset-function-examples.html
`rstc_set_external_reset(RSTC, 13);	/* (2^(13+1))/32768 */ `
delay is 0.5 sec here, so 256 is too big.

fredlcore:
Using both lines and a value of 13 also does not work :(...

fredlcore:
So far 
`rstc_start_software_reset(RSTC);`
is the only one that works...

dukess:
Ok. We will use it until board arrive to me.

PR #117, opened 2020-07-11
binary search in findLine(), _printFIXPOINT() in MQTT

dukess:
binary search in findLine(). It must be more faster than old search (11-18 iterations instead 1-129), _printFIXPOINT() in MQTT
small optimizations
add/replace some russian texts


fredlcore:
Thanks, could you also fix one bug regarding HIDE_UNKNOWN: The first one is fixed already, it should have been "return" instead of "continue", but still no "error 7" parameters are printed out (in grey) as before. Somehow these "error telegrams" are filtered before they get to query_printHTML. Could you check that please? Thanks!

dukess:
Tried to fix it. Unfortunately without testing because i in business trip.

dukess:
@fredlcore Fixed and tested.

fredlcore:
Thanks!

PR #120, opened 2020-07-21
Update BSB_lan.ino

dukess:
sprintf_P in Due is a macros.
This macors can't be correctly used without arguments.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -1208,7 +1208,7 @@ void loadPrognrElementsFromTable(int i){
   decodedTelegram.precision=pgm_read_byte_far(pgm_get_far_address(optbl[0].precision) + decodedTelegram.type * sizeof(optbl[0]));
   decodedTelegram.unit_len=pgm_read_byte_far(pgm_get_far_address(optbl[0].unit_len) + decodedTelegram.type * sizeof(optbl[0]));
   #if defined(__SAM3X8E__)
-  memcpy(decodedTelegram.unit, optbl[decodedTelegram.type].unit, div_unit_len);
+  memcpy(decodedTelegram.unit, optbl[decodedTelegram.type].unit, decodedTelegram.unit_len);
   #else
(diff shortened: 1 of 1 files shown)


fredlcore:
Thank you for this quick fix, it compiles now without errors and I have merged this for now so that we have a working master branch, but I think this solution is harder to maintain in the future because one always has to keep in mind to update the outBufLen command at the end so that it matches the string length. And errors will be difficult to spot because they only become visible much later when outBuf is being printed.

I really would suggest to revert back to the standard sprintf command and not use sprintf_P in cases where it does not work seamlessly on the Due. The few bytes that are being saved for the Mega (and only for the Mega because the Due does not have separate PROGMEM flash memory) are not worth the less convenient code. And since the switch to the Due, code should be optimized for the Due and not for the Mega (although we should try to maintain compatibility for the Mega as long as possible).

dukess:
I don't like this solution either, so I'll think of ways to make it more 
beautiful and readable

fredlcore <notifications@github.com> 21 июля 2020 г. 18:12:23 написал:

fredlcore:
Great, thanks a lot!

dukess:
Just FYI:
I try to redefine sprintf_P and it compile correct 

```
#if defined(__SAM3X8E__)
#ifdef sprintf_P
#undef sprintf_P
#endif
#define sprintf_P(s, ...) sprintf((s), __VA_ARGS__)
#endif
```
We can rollback and using this version sprintf_P instead strcpy + manual string length count.

I have an idea how to do output buffering with automatic sending to the client, but the overhead will become unacceptable: either we need to count bytes before writing to the buffer (CPU cycles),or make a buffer with a large amount of memory (~1500 bytes).

fredlcore:
Oh, very good, thanks!
As for the output buffering, I'm not sure I understand the scope, but at least on the Due, 1.5kB is not really much...

dukess:
I always forget about 96 KB of memory :)

PR #121, opened 2020-07-22
Update BSB_lan.ino

dukess:
sprintf_P() was redefined for Arduino Due.
strcpy_P() was reverted back to sprintf_P()

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -513,6 +513,14 @@ uint8_t json_types[20] = { 0 };
   dht DHT;
 #endif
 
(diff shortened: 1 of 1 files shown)


fredlcore:
Hi Sergey,

ah, glad it arrived safely! 
As for powering the board, you can use the 12V and possibly even go directly into the DC connector, but we don't know how much power can be drawn from that line, so we don't officially mention it. Also, this would make the galvanic separation between bus and Arduino null and void. Therefore, we don't recommend it.

Cheers,


Frederik

fredlcore:
Thanks!

PR #122, opened 2020-07-24
Buffered print to web client with automatic flushing.

dukess:
Buffered print to web client with automatic flushing.
transmitFile() use bigBuff now instead dynamic allocation.
Code preparation for compiling for modern MCU.


fredlcore:
Thanks!

PR #124, opened 2020-07-26
Update BSB_lan.ino

dukess:
Bugfix: /Q command was shown Programs with VT_UNKNOWN type (Error 260) instead UNKNOWN telegrams (Error 257).
Bugfix: Programs with VT_UNKNOWN type not shown raw data in JSON format

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -3796,7 +3796,7 @@ char *build_pvalstr(boolean extended){
   strcpy_P(outBuf + len, PSTR(":"));
   len+=strlen(outBuf + len);
   }
-if(decodedTelegram.value[0] != 0 && decodedTelegram.value != 260){
+if(decodedTelegram.value[0] != 0 && decodedTelegram.error != 260){
   strcpy_P(outBuf + len, PSTR(" "));
   strcat(outBuf + len, decodedTelegram.value);
   len+=strlen(outBuf + len);
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -2697,6 +2697,10 @@ void printTelegram(byte* msg, int query_line) {
             case VT_UNKNOWN:
             default:
               prepareToPrintHumanReadableTelegram(msg, data_len, bus.getPl_start());
+              int len = 0;
+              for (int i=0; i < data_len; i++) {
+                len += sprintf_P(decodedTelegram.value + len,PSTR("%02X"),msg[bus.getPl_start()+i]);
+              }
               decodedTelegram.error = 260;
               break;
           }
@@ -3792,7 +3796,7 @@ char *build_pvalstr(boolean extended){
   strcpy_P(outBuf + len, PSTR(":"));
   len+=strlen(outBuf + len);
   }
-if(decodedTelegram.value[0] != 0){
+if(decodedTelegram.value[0] != 0 && decodedTelegram.value != 260){
   strcpy_P(outBuf + len, PSTR(" "));
   strcat(outBuf + len, decodedTelegram.value);
   len+=strlen(outBuf + len);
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks!

dukess:
You're welcome!

fredlcore:
Hm, something still does not seem right - when I comment out one of the parameters for my heater (device family 97), for example this line:
```
{0x053D2FF1,  CAT_FEHLER,           VT_BYTE,          6800,  STR6800_2,0,                    NULL,         FL_RONLY,     DEV_097_ALL}, // [ ] - THISION: Fehler Historie 1
```
/Q should report this command ID because my heater knows 053D2FF1, but not 053D06D3 which is the DEV_ALL command ID.
However, /Q returns with an empty test.

The idea of /Q is to go through all command IDs which are not DEV_ALL command IDs and if they return an answer message type (0x07), we have a hit, because then one of the other, machine specific command IDs matches this machine as well.

I hope this error has not lasted for a longer time already, because I was wondering that for the last few weeks/months, all the /Q reports either returned empty or just with the 5956/6050 parameters...

fredlcore:
This is what shows up in the debug console:
```
LAN->HEIZ QUR      053D2FF1                                                     
DC C2 00 0B 06 3D 05 2F F1 9B 2C                                                
HEIZ->LAN ANS      053D2FF1 00 0F                                               
DC 80 42 0D 07 05 3D 2F F1 00 0F F1 99                                          
LAN->HEIZ QUR 6800 Fehler - Historie 1 Datum/Zeit:                              
DC C2 00 0B 06 3D 05 06 D3 23 72                                                
HEIZ->LAN ERR 6800 Fehler - Historie 1 Datum/Zeit: error 7                      
DC 80 42 0C 08 05 3D 06 D3 07 C8 20                                             
#6800:  (parameter not supported)      
```
So the first test with 053D2FF1 returns message type 0x07 and the second test with the DEV_ALL command id (05 3D 06 D3) returns message type 0x08 (error). But the first test should nevertheless be reported...

fredlcore:
When I set the second decodedTelegram.error condition from == 257 to == 0, it shows results, but also those where already DEV_ALL is working. The original idea and functionality was that only those command IDs are shown where a machine specific command ID matches, but NOT the DEV_ALL command ID. Because sometimes a machine specific command ID matches, but also the DEV_ALL one, and then we don't need the output.

fredlcore:
Ok, managed to fix it myself :)...

dukess:
I thought I'd run the same test tonight, but you were faster. Thanks again for describing the algorithm - it can help a lot with debugging.

dukess:
It turns out that my error lay in my incorrect interpretation of the zero length of the string "pvalstring" when I entered numeric error codes.

fredlcore:
Well, it took me a while to understand again what I did there as well ;) - but with your code it is much cleaner and easier to understand! 

One last thing: could you turn off the buffered output for /Q? Since there are often long pauses between outputs, it may seem as if the Arduino hangs, at the beginning even the menu bar is not fully printed.

dukess:
>much cleaner and easier to understand!

I'm glad it helped, but if I hadn't changed the code, it wouldn't have been necessary. 🤣 

>could you turn off the buffered output for /Q? 

I can't completely disable buffering without rewriting /Q, but I can add more calls flushToWebClient() function.

fredlcore:
Well, it happens to me all the time - the most important thing is we figured it out, and it was not a critical function. I narrowed it down to only four people who were affected and I can simply ask them again, no problem.

Rewriting /Q is of course not an option, but I tried adding it after the "Start Test..." message, and it still is not showing up right away, it pauses before the last two meno items are displayed. Also, I tought that in the current code, it would display at least after each "hit", but that is also not the case. It seems that a minimum amount of bytes has to be filled in order for that part of the page being displayed. Or is this browser-specific?

PR #125, opened 2020-07-27
Reducing intervals between buffer flushes to web client.

dukess:


fredlcore:
Thanks, will try it out!

fredlcore:
Hm, it's a bit better, but the main problem remains: When starting the function, the output stops before the last two menu items. This is because there seems to be too few data before the device scan begins. This takes around 10 seconds until the next output is prepared. During that time the output seems to hang. Impatient people could think that something is wrong. If only the "Start Scan..." message would already be displayed and not just a "broken" menu, that would certainly help and we would not need all the flush commands between the first fixed parameters...

dukess:
It depend from buffer size. It size is between 900 and 1350 bytes (more than 2 * OUTBUF_SIZE and less then 3 * OUTBUF_SIZE). It looks like sending the buffer over the network is blocked while the query() is running. So we can try to force flush it.

PR #127, opened 2020-07-29
New paraeters from Baxi Luna Platinum+ (Heater and Sitherm Pro categories).

dukess:
Bugfix: VT_UINT100 decoded as float now
VT_TEMP_SHORT_US type added for some parameters
/Q: code for compatibility with Mega added
New paraeters from Baxi Luna Platinum+ (Heater and Sitherm Pro categories).
Category CAT_FEHLER was moved to cmdtbl2.

Frederik, look at these parameters please: 2662, 2663


fredlcore:
Great, thanks! Just a one question: the AVR code for /Q would work on the Due as well, wouldn't it? Then it's probably easier to read if we just keep the Mega code?
As for 2662/2663, you are right, it's minutes, not seconds - it was even sent to me correctly, don't know what happened that I ended up with seconds :)...

dukess:
>the AVR code for /Q would work on the Due as well, wouldn't it?
This code will work on both family, so it's up to you. :-)

fredlcore:
Ok, thanks, I've just removed the Due-specific code then. Won't hurt the Due to have a few more lines :)...

PR #128, opened 2020-07-29
RVS61.843E controller

LucekDev:
Added RVS61.843E controller. New/corrected commands for the controller in the range "Trinkwasserspecher" and "Trinkwasser".


fredlcore:
Thanks! Just one question: Is there a reason why you used DEV_108_217 on all parameters you added?
The general rule is this:
- If it is a completely new command, use DEV_ALL (because there is no other command ID to differentiate from)
- If a DEV_ALL already exists, but your command ID from your machine differs, use DEV_XXX_ALL with XXX being the device family, in your case: DEV_108_ALL.
- if a DEV_XXX_ALL already exists for your device family, use DEV_XXX_YYY with YYY being the device variant. In your case: DEV_108_217.

Could you please update your added parameters in that way? Thank you!

LucekDev:
Done. 
Is there a way to disable/hide commands not present in the controller?

fredlcore:
Yes, use the #define HIDE_UNKNOWN configuration in _config.h.
Two more things I noticed:
For new ENUMs and ENUM-Options, could you check whether the text for this new parameter/option is already defined?
I just saw that options such as "Keine" or "Aus" already exist at some other place (for example as MENU_TEXT_OFF). But also other, more specific options may already be defined somewhere else. This is important for translations because then people don't have to translate the same word twice and it will automatically be displayed in the correct translation.

The second thing is whether you could check if one of the new parameters you found has the same command ID as the ones listed from 10200 onwards. You can just do a text search for the new command ID (it should not appear anywhere else), but if it is already listed in the 10200++ section, you can just remove that line (which saves 17 bytes and cleans up the list of unknown parameters).

Thank you!

fredlcore:
Also, I just saw that command ID 0x253D1158 which you added for 1601 is already used for 1600, albeit for a specific device family. Does it make more sense to add your device family to 1600 as well or is it really "manual push" that that command ID does for you? At least from the description, it seems to fit better to 1600.

fredlcore:
I did the comparison with the 10200++ parameters for you, it's only one that is affected and can be removed: 10519 (0x313D0757)

LucekDev:
(0x313D0571) 1600 Betriebsart [Aus, Ein, Eco]
(0x253D1158) 1601 Betriebsartwahl Eco [Keine, Trinkwasserspeicher]

In Browser I see:
1600 Trinkwasser - Trinkwasserbetrieb: 1 - Ein
1601 Trinkwasser - Betriebsartwahl Eco: 0 - Keine
 so for mee both commands work (omitting deviation in text 1600). 1601 tested with both values is ok.

fredlcore:
The thing is that 1601 is "DHW Push" which is something different than "DHW mode". You don't see DHW Push because your device family overrides the general parameter.
If your command ID also changes the mode and does not initiate a push,I would add your command ID and device family under 1600 and - if necessary - add a new ENUM for your device family.

fredlcore:
I think we're not on the same page: Now you have added this line:
```
{0x313D0571,  CAT_TW,               VT_ENUM,          1600,  STR1600_2,sizeof(ENUM1600),     ENUM1600,     DEFAULT_FLAG, DEV_108_ALL}, // [-] - Trinkwasser - Betriebsart [Aus, Ein, Eco]
```
which is the same as the one for DEV_ALL for 1600, so it is superfluous.

On the other hand, you still have this line:
```
{0x253D1158,  CAT_TW,               VT_ENUM,          1601,  STR1601_2,sizeof(ENUM1601_2),   ENUM1601_2,   DEFAULT_FLAG, DEV_108_ALL}, // [-] - Trinkwasser - Betriebsartwahl Eco: Keine, Trinkwasserspeicher
```
What does this command ID do for you? Does it change the mode of DHW ("Trinkwasser")? Or does it initiate a DHW Push (heat up drinking water even if the minimum temperature has not been reached)?

My safe bet is that you are changing the DHW mode with this. Then you don't add it under 1601 (which is DHW push) but under 1600 (which is DHW mode). You may have to add another ENUM (like ENUM_1600_2) if the parameter options do not match your heater. 

But even if this command ID does a DHW push for you, there would not be an ENUM saying "Trinkwasserspeicher", because it is a On/Off parameter.

Please change this accordingly and I can accept the commit. Please also remove the file localization/www/lang_de.js.gz because it prevents your commit from being merged. Thanks.

LucekDev:
With line for 1600 (DEV_108_ALL) I change text: STR1600_2 to be as on display of ma controller. With current code for both 1600 & 1601 it will be exactlly as is presented on the display. In my controller there are two parameters: 1600 & 1601.

If you search in Google "CE1U2355de_052", then you should find PDF: "Albatros Wärmepumpenregler Benutzerhandbuch RVS61.843" with description of parameters - it fits quite well with parameters in my controller.

1600 Betriebsart: [Aus, Ein, Eco]
1601 Betriebsartwahl Eco: [Keine, Trinkwasserspeicher]

**Betriebsart**
Über die "Betriebsart" kann die Trinkwasserladung `Ein`, `Aus `oder auf `Eco`-Betrieb
geschaltet werden.

Werden grössere Temperaturschwankungen beim Trinkwarmwassers akzeptiert
und steht Energie über Solarheizung oder Feststoffkessel frei zur Verfügung, kann
die Eco-Funktion verwendet werden. Falls lokale Vorschriften bezüglich
Legionellenfunktion bestehen, müssen diese beachtet werden.

**Betriebsartwahl Eco**
`Keine`
Die Auswahlmöglichkeit "Eco" bei "Betriebsart" ist ausgeblendet.
`Trinkwasserspeicher`
Der Eco-Betrieb wird für den Trinkwasserspeicher verwendet.
Im Eco-Betrieb ist die Trinkwasser-Erwärmung durch steuerbare Erzeuger
eingeschränkt. Diese werden nur eingeschaltet, wenn das TWW-Reduziertniveau
unterschritten wird oder bei aktiver Legionellenfunktion.

Der manuelle Push kann auch im Eco-Betrieb aktiviert werden

fredlcore:
Ah, ok, I'll check it and see how we can deal with the existing 1601, because your heater may still support a push, and then you would lose that function if 1601 has a different meaning in your heater. I'll get back to you later...

LucekDev:
By the way I updated yesterday to 1.0.32-20200728190440 and it seems that the logging is not working. Before with 1.0.23-20200714123325 it worked without any problems. Now after /D0 I could see header, then after 1 min (log each 60 s) first record and then next trays nothing.

fredlcore:
@dukess : could this have to do with your changes?

fredlcore:
@1coderookie : Do you have an idea how we should deal with the 1600/1601 issue?

LucekDev:
I did not touch BSB_lan.ino (and no custom code), so I don't think - and my changes worked with 1.0.23

fredlcore:
No, I was asking @dukess

dukess:
I'll watch it later tonight.

fredlcore <notifications@github.com> 30 июля 2020 г. 17:19:34 написал:

dukess:
Can you show /C output?
At first glance, there are no problems, but it is better to compare the configurations.

LucekDev:
Curently I installed mix of version 1.023 and all changed files with exception of BSB_lan.ino. Wiht this setup /D /DG are working fine. Yesterday I switched off IPWE, so my first thought was that it is a kind of side effect, but today I went back to the following configuration. Independly of IPWE with 1.0.32-20200728190440 there was no graph /DG or no data /D. After /D0 and then /D I could see:
- header
- after 1 min first set of data
- after next min: no header, no data (no graph /DG)
Browser: Firefox 79.0 on PC, and Safari on tablet.

> Bus-System: BSB (66, 0) Nur Lesen
> Monitor Modus: 0
> Verbositäts-Level: 1
> 
> Installierte Module:
> IPWE, LOGGER, VERSION_CHECK
> 
> free space: 1903 MB
> free clusters: 487328
> freeClusterCount() call time: 2 microseconds
> 
> Berechnung von 24h-Mittelwerten für die folgenden Parameter:
> 8700 - Außentemperatur
> 
> Loggen der folgenden Parameter alle 60 Sekunden:
> 8700 - Außentemperatur
> 8743 - Vorlauftemperatur 1
> 8410 - Rücklauftemperatur WP
> 8412 - Vorlauftemperatur WP
> 8415 - Heissgastemperatur 1
> 8427 - Quelle Eintrittstemperatur
> 8429 - Quelle Austrittstemperatur
> 8830 - Trinkwassertemperatur 1
> 8832 - Trinkwassertemperatur 2

fredlcore:
@1coderookie : As for parameter 1601 which seems to be officially "Betriebsartwahl Eco", I suggest we move the manual push from 1601 to 1603. That leaves at least 1602 ("Status Trinkwasserbereitung" / DHW Status) in its current place. Users who use manual push will have to adjust their parameter, unfortunately, but I think it's better to stick with the official parameter numbers rather than moving stuff around...

dukess:
I tried to reproduce the problem but without any result:
```
Milliseconds;Date;Parameter;Description;Value;Unit
331713;30.07.2020 22:48:15;8700;STR8700_TEXT;9.9;UNIT_DEG_TEXT
331878;30.07.2020 22:48:15;8314;STR8314_TEXT;43.8;UNIT_DEG_TEXT
332053;30.07.2020 22:48:15;8830;STR8830_TEXT;37.8;UNIT_DEG_TEXT
352258;30.07.2020 22:48:35;8700;STR8700_TEXT;9.9;UNIT_DEG_TEXT
352432;30.07.2020 22:48:36;8314;STR8314_TEXT;43.7;UNIT_DEG_TEXT
352647;30.07.2020 22:48:36;8830;STR8830_TEXT;37.8;UNIT_DEG_TEXT
372819;30.07.2020 22:48:56;8700;STR8700_TEXT;9.9;UNIT_DEG_TEXT
373033;30.07.2020 22:48:56;8314;STR8314_TEXT;43.7;UNIT_DEG_TEXT
373199;30.07.2020 22:48:56;8830;STR8830_TEXT;37.8;UNIT_DEG_TEXT
```

dukess:
By the way...
log file for telegrams and parameters was splitted some time ago, and now telegrams saved to "journal.txt". It can be readed with /DJ command, created with /DJ0 (/D0 recreated both files). Telegram logging can be enabled with /LE command. /LB=1 log broadcast only, /LU=1 - unknown only.

fredlcore:
Just flashed the most recent version on my Due and logged three parameters every 60 seconds, works without problems, displays the CSV as well as the graph with /DG.

fredlcore:
I accepted your pull request now and moved DHW push from 1601 to 1603, please check if the parameters now work as you intend them to, thanks!

LucekDev:
Parameters are working as expected. 

/D & /DG is in my case still not working, event with 1.0.38-20200730234859. I checked this on few different systems / browsers with the same result. I checked that in case of /D I get 60 bytes as response: code 200 OK, Conetent-Type: text/plain; charset=utf-8 and no more data.
I am using non standard port - 6123 instead of 80, but it cannot influence only /D /DG or?

fredlcore:
That's strange, works on my system with 1.0.36-20200729174905 both with /D and /DG and this config:
```
Version: 1.0.36-20200729174905
Freier Speicher: 85439 Bytes
Uptime: 61127219
Bus-System: BSB (66, 0) Lesen/Schreiben
Monitor Modus: 0
Verbositäts-Level: 4
1-Wire Bus Pins: 7
Gesperrte GPIO Pins: 10 11 12 13 50 51 52 53 62 63 64 65 66 67 68 69
MAC-Adresse: 00:80:41:19:69:90

Installierte Module:
IPWE, RESET, MQTT, DEBUG, LOGGER, VERSION_CHECK

Free space: 15273 MB
free clusters: 1955029
freeClusterCount() call time: 2 microseconds

Berechnung von 24h-Mittelwerten für die folgenden Parameter:
8700 - Außentemperatur
8323 - Gebläsedrehzahl
8326 - Brennermodulation

Loggen der folgenden Parameter alle 60 Sekunden:
8700 - Außentemperatur
8743 - Vorlauftemperatur 1
8314 - Kesselrücklauftemperatur Ist
```

Just to rule out that it is the different port (which should not be an issue), could you try it with port 80?

LucekDev:
I tried with port 80 - no change still /D & /DG are not working.

fredlcore:
Very strange. Can you try with just the parameters I'm using? And can you read the contents of the file on a PC and see if the data is actually written or if already writing the data is a probllem?

PR #129, opened 2020-07-30
New parameters from Baxi Luna Platinum+ (HK1 and Trinkwasserspeicher)

dukess:
Bugfix: JSON generation error (not enough ',')

const char ENUM6706[] PROGMEM_LATEST:

"\x14 " ENUM6706_14_TEXT "\0"
"\x15 " ENUM6706_15_TEXT "\0"
"\x16 " ENUM6706_16_TEXT "\0"
was replaced with
"\x0e " ENUM6706_14_TEXT "\0"
"\x0f " ENUM6706_15_TEXT "\0"
"\x10 " ENUM6706_16_TEXT "\0"
it's correct?


fredlcore:
Yes, I think the ENUMs are correct the way you changed it.
As for parameter 794, command ID 2D3D0609 returns a valid answer (although it seems to be a DWORD type) for my machine, whereas the one you found (2D3D0639) returns an error. I would therefore recommend we leave 2D3D0609 in place with DEV_ALL (because I don't remember from which machine it came from) and add your 2D3D0639 for DEV_123_ALL.

As for all the other command IDs you found, did you have time to check against the brute force parameters from 10200++ onwards? If the command IDs are listed in that section already, you could remove the doublettes and save more flash memory. Thanks!

dukess:
Each command I added is contained only once, I checked it with a search. If I understand you correctly. Or is there some other procedure for checking for duplicates?

fredlcore:
Hi again, I just checked and I think that 14/15/16 are correct, or do you have any indication why it is not correct?

fredlcore:
I think I got them from the Siemens software back then but I can't check it right now...

fredlcore:
As for duplicates, yes, that's what I meant, thank you!

dukess:
My heater show code 14 (DEC) and 16 (DEC), so I thought there was a typo in ENUM. Unfortunately, the heater does not output any description in addition to the code, so you can only guess about the possible options.

dukess:
I revert changing in ENUM6706.

fredlcore:
I remember where I found them, here on page 40:
https://polo.broetje.de/pdf/sgb2_320_ih_10_395319.pdf
(sorry, it's in German only).
Since the other numbers (such as 99 for "Störstellung") are corresponding with the return values in the telegrams, I assume that 20 to 22 are also correct. If your heater shows 14 to 16, maybe these are different phases? 22 (0x16) would be "Außerbetriebsetzung" (something like turning off), that shouldn't happen very often, I think, so if 16 occurs on your heater frequently, it may mean something else?

fredlcore:
I saw you added some new parameters with CMD_UNKNOWN, may I ask how you know the data type if you don't know  the command ID / telegram?

dukess:
Thank you for new info. i read it little bit later. 
About CMD_UNKNOWN: еhese are the same programs, but for two other heating circuits. I know their prognr and format (with a high probability). However, since I only have one heating circuit, I will not be able to quickly find out the appropriate commands yet.

fredlcore:
Ah, ok, understand - as for other heating circuits, it is often the first byte that is increased by 1, so 05/21/25/2D/31 is often heating circuit 1, 06/22/26/2E/32 is heating circuit 2 etc. You can relatively safely check that by querying the parameter with the calculated command ID and you should get a response in the same data format.

dukess:
Oh, thanks!

PR #130, opened 2020-08-01
New parameters was copied from HC1 to HC2 and HC3/P

dukess:
And tested, of course


fredlcore:
Great, thanks! I saw that some of the new parameters are for DEV_123_ALL only. As a general rule, all new parameters should bei DEV_ALL, because it may be the only command ID available and thus fits all heaters where it's accessible. And it won't hurt the others...

fredlcore:
And for 1182, "DEV_123_ALL" would be enough. "DEV_123_231" would only be necessary if there was a "DEV_123_ALL" already with a different command ID that is not working for your heater.

dukess:
I left the device ID unchanged for cases when there is already definition for all devices or 123 family in the HC1. Does it make sense to copy the existing definitions from HC1 to others HC's?

fredlcore:
Ok, but what about these lines:
```
{0x2E3D0639,  CAT_HK2,              VT_GRADIENT,      1094,  STR1094,  0,                    NULL,         DEFAULT_FLAG, DEV_123_ALL}, // [min/K ] - Heizkreis 1 - Aufheizgradient

{0x223D04AB,  CAT_HK2,              VT_PERCENT,       1182,  STR1182,  0,                    NULL,         DEFAULT_FLAG, DEV_123_231}, // [%] - Heizkreis 1 - Pumpendrehzahl Minimum

{0x063D10E1,  CAT_HK2,              VT_PERCENT,       1185,  STR1185,  0,                    NULL,         DEFAULT_FLAG, DEV_123_ALL}, // Baxi Luna Platinum+ Pumpe-PWM Minimum OEM [%]
{0x063D10E0,  CAT_HK2,              VT_PERCENT,       1186,  STR1186_2,0,                    NULL,         FL_RONLY, DEV_123_ALL}, // Baxi Luna Platinum+ Pumpe-PWM Maximum OEM [%]

{0x2F3D0639,  CAT_HKP,              VT_GRADIENT,      1394,  STR1394,  0,                    NULL,         DEFAULT_FLAG, DEV_123_ALL}, // [min/K ] - Heizkreis 1 - Aufheizgradient

{0x233D04AB,  CAT_HKP,              VT_PERCENT,       1482,  STR1482,  0,                    NULL,         DEFAULT_FLAG, DEV_123_231}, // [%] - Heizkreis 1 - Pumpendrehzahl Minimum

{0x073D10E1,  CAT_HKP,              VT_PERCENT,       1485,  STR1485,  0,                    NULL,         DEFAULT_FLAG, DEV_123_ALL}, // Baxi Luna Platinum+ Pumpe-PWM Minimum OEM [%]
{0x073D10E0,  CAT_HKP,              VT_PERCENT,       1486,  STR1486_2,0,                    NULL,         FL_RONLY, DEV_123_ALL}, // Baxi Luna Platinum+ Pumpe-PWM Maximum OEM [%]


```
There is no other entry for parameter 1094, 1182,... etc., so it should only be DEV_ALL, shouldn't it?

dukess:
I'm thinking of copying the "missing" lines from HC1 to HC2/3.
Ex:
`{0x2D3D0609,  CAT_HK1,              VT_GRADIENT,      794,   STR794,   0,                    NULL,         DEFAULT_FLAG, DEV_ALL}, // [min/K ] - Heizkreis 1 - Aufheizgradient`
to 
`{0x2E3D0609,  CAT_HK2,              VT_GRADIENT,      1094,   STR1094,   0,                    NULL,         DEFAULT_FLAG, DEV_ALL}, // [min/K ] - Heizkreis 2 - Aufheizgradient`
`{0x2F3D0609,  CAT_HKP,              VT_GRADIENT,      1394,   STR1094,   0,                    NULL,         DEFAULT_FLAG, DEV_ALL}, // [min/K ] - Heizkreis 3/P - Aufheizgradient`
Or this is bad idea?

fredlcore:
No, that most likely makes sense, if you get a similar result for these parameters (sometimes this works even if you don't have that many HCs).
My question was why you used DEV_123_ALL etc. in the lines quoted above and not DEV_ALL?

dukess:
First circuit already has a lines with DEV_ALL, and if i write my parameters for HC2/3 as DEV_ALL instead DEV_123_ALL,  the behavior of these parameters will change in comparsion to HC1.
So I ask if it is possible to painlessly copy existing parameters with DEV_ALL from HC1 to HC2/3 (794 to 1094 and 1394 as exapmle).

fredlcore:
I'm not sure if I understand correctly, but if you copy the lines to new parameter numbers in HC2 etc. and adjust the command ID, these parameters will have nothing to do with the HC1 parameters. So it should be safe to use DEV_ALL. Or do you mean something else?

dukess:
Classic case "lost in translation" :-)
 I made patch with changes that i mean.

fredlcore:
Ok, thanks - just one last question: Did you check if the "guessed" parameters in HC2/3/P return some kind of valid results? I think for 05/21/2d and the others I mentioned, this is rather safe to assume that you can just increase the first byte, but for 11 and 19 I'm not so sure. I tried it with 193D2F88 (which works for me) and with 1A3D2F88 (which does not), so unless you get a valid-looking response, I would leave those ones out. Also check if the response actually contains the same command ID as the query telegram because otherwise it could just be a broadcast telegram that is received while waiting for an answer.

dukess:
I understand your concern.
I tested the parameters that i added for my heater, they all return data very similar to the correct ones.
Other "guessed" options i have, most likely, will not work (but i not tested them). Maybe then comment them out and enter CMD_UNKNOWN instead of the command ID and leaving them as reminder? Or is it better to do as you said above: put DEV_ALL instead of DEV_123_ALL?

dukess:
By the way, yesterday I found two small hidden categories that appear only when the input function H4 or H5 is set as "Verbrauchanfonderung VK1" or "Verbrauchanfonderung VK2" or " Freigabe Swimmbad Erzeuger".

Can i add these categories to list?
What i should to do with matching programs (1859 same as 5952, etc)?


Category: Потребитель контур 1 | Verbraucherkreis 1
Prognr: 1859 same as 5952
Name: Требуемая постоянная уставка температуры подачи | Vorlaufsollwert Verbr'anfo

Prognr: 1874
Name: Приоритет смены ГВС | TWW-Ladevorrag
Value: 255  yes
Unit: ENUM_YESNO

RGT1->HEIZ QUR      293D08E5
DC 86 00 0B 06 3D 29 08 E5 5C F2
HEIZ->RGT1 ANS      293D08E5 00 FF
DC 80 06 0D 07 29 3D 08 E5 00 FF 1A 9B

Value: 0  no
Unit: ENUM_YESNO
RGT1->HEIZ SET      293D08E5 01 00
DC 86 00 0D 03 3D 29 08 E5 01 00 4E D2
HEIZ->RGT1 ACK      293D08E5
DC 80 06 0B 04 29 3D 08 E5 77 6A
RGT1->HEIZ QUR      293D08E5
DC 86 00 0B 06 3D 29 08 E5 5C F2
HEIZ->RGT1 ANS      293D08E5 00 00
DC 80 06 0D 07 29 3D 08 E5 00 00 04 6B

Prognr: 1875
Name: Чрезмерный наддув тепла | Ubertemperaturabnahme
Value: 1 on
Unit: ENUM_ONOFF
RGT1->HEIZ QUR      293D08CA
DC 86 00 0B 06 3D 29 08 CA 89 7F
HEIZ->RGT1 ANS      293D08CA 00 01
DC 80 06 0D 07 29 3D 08 CA 00 01 BE BD

Value: 0 off
Unit: ENUM_ONOFF
RGT1->HEIZ SET      293D08CA 01 00
DC 86 00 0D 03 3D 29 08 CA 01 00 E4 25
HEIZ->RGT1 ACK      293D08CA
DC 80 06 0B 04 29 3D 08 CA A2 E7
RGT1->HEIZ QUR      293D08CA
DC 86 00 0B 06 3D 29 08 CA 89 7F
HEIZ->RGT1 ANS      293D08CA 00 00
DC 80 06 0D 07 29 3D 08 CA 00 00 AE 9C

Prognr: 1878 same as 6130
Name: С буферным накопительным баком  | Mit Pufferspeicher

Prognr: 1880 same as 10453
Name: С основным контроллером/насосом системы | Mit Vorregler/Zubring'pumpe
Value: 255  yes
Unit: ENUM_YESNO


Category: Потребитель контур 2 | Verbraucherkreis 1
Prognr: 1909 same as 6048
Name: Требуемая постоянная уставка температуры подачи |  | Vorlaufsollwert Verbr'anfo


Prognr: 1924
Name: Приоритет смены ГВС  | TWW-Ladevorrag
Value: 255  yes
Unit: ENUM_YESNO
RGT1->HEIZ QUR      2A3D08E5
DC 86 00 0B 06 3D 2A 08 E5 05 A2
HEIZ->RGT1 ANS      2A3D08E5 00 FF
DC 80 06 0D 07 2A 3D 08 E5 00 FF D4 7B


Prognr: 1925
Name: Чрезмерный наддув тепла | Ubertemperaturabnahme
Value: 1 on
Unit: ENUM_ONOFF
RGT1->HEIZ QUR      2A3D08CA
DC 86 00 0B 06 3D 2A 08 CA D0 2F
HEIZ->RGT1 ANS      2A3D08CA 00 01
DC 80 06 0D 07 2A 3D 08 CA 00 01 70 5D

Prognr: 1928 same as 6134
Name: С буферным накопительным баком | Mit Pufferspeicher

Prognr: 1930 same as 10457
Name: С основным контроллером/насосом системы  | Mit Vorregler/Zubring'pumpe
Value: 255  yes
Unit: ENUM_YESNO

fredlcore:
Just to be really sure: 0x1A3D2F88 returned some kind of valid data for you? And the command ID in the return code is the same as in the query telegram (with only the first two bytes of command ID switched, as usual)? I'm just asking because when I tried this, the return telegram was actually from a broadcast telegram.

But if this is correct with 1A3D2F88 and the others, then we should add all of them in the way you did and I will merge the pull request as it is. Just want to be really sure because so far we have never had any command IDs starting with 1A...

fredlcore:
> By the way, yesterday I found two small hidden categories that appear only when the input function H4 or H5 is set as "Verbrauchanfonderung VK1" or "Verbrauchanfonderung VK2" or " Freigabe Swimmbad Erzeuger".

Wow, that is amazing! Yes, please add these two categories? And with matching parameters, you can just use those ENUMs etc., so translators don't have to translate twice. Once 10453 and 10457 have been included as "normal" parameters, you can remove them from 10xxx. Thanks a lot!

fredlcore:
Actually, I found these parameters in this document:
https://polo.broetje.de/pdf/c0026021gema_ersatz_bsw_b-d_rvs61_e_7622600-01.pdf
on page 63, maybe this is helpful for you?
@1coderookie : Did you know about these?

fredlcore:
Since there are quite a few new categories (also Vorregler/Zubringerpumpe 2, I haven't checked all of them in the document), we should add all of them now so that users will only have to adjust their categories once).

fredlcore:
Ok, I will add all the new categories in this document now...

fredlcore:
Added...

dukess:
>Just to be really sure: 0x1A3D2F88 returned some kind of valid data for you? 

No, i not tested it. Other commands added in last commit not tested too.  But all commands that was added in this commit https://github.com/fredlcore/bsb_lan/pull/130/commits/67a9f6a69c0202c2ee2325736d69a7954beabc41 was tested and shown me valid data.

fredlcore:
Ok, could you then add those ones that returned valida data with the command ID and the others with CMD_UNKNOWN? Then I can merge the pull request.

1coderookie:
> @1coderookie : Did you know about these?
What do you mean? The new categories and parameters (like "Verbraucherkreis" and so on)? Yes, I mentioned it already approx. 2years ago when I stumbled across them at a Brötje BOB with the new RVS43.325. Also in the FHEM-forum I mentioned it a few times, put them down in a list, posted it and asked every new BOB-users to decode them.. ;)

fredlcore:
Ok, then it must have slipped my attention because I should have at least added/reserved the category names... Well, back then or now, in the end it doesn't make that much of a difference. An no one of them has made the effort to decode them? Shame on them ;)!

dukess:
> And with matching parameters, you can just use those ENUMs 

Can i rename and move 5952 to 1859, 6130 to 1878, etc?

fredlcore:
I mean something like this:
`#define STR1859 STR1859`
and the same for ENUMs etc.

dukess:
> I mean something like this:
#define STR1859 STR5952 
and the same for ENUMs etc.

Sure i do that.

I mean program lines, here is three variants: 
1. leaving 5952 without changing, nothing add to 1859
2. Copying 5952 to 1859 (two lines with same command code)
3. Moving 5952 to 1859
What i must do?

dukess:
> An no one of them has made the effort to decode them? Shame on them ;)!

Perhaps no one needs these parameters. 🤣

fredlcore:
Ah, now I get it :)! It's exactly the same command ID? Then you can just copy it, because maybe in the future there will be different command IDs for different heaters, so it's better to have it in both locations. If any of the texts is different, you can add a new STR etc. for it. Thanks!

dukess:
Roger that!

dukess:
Done!

fredlcore:
Thanks! Hopefully one last thing: you now added two lines for some parameters (I think 1182 and 1184 for example), one unknown for all and one with command id for 123. If the one for 123 works for you and there is no other command id, then set it to DEV_ALL. Only those command ids that do not work for you should be added with Unknown.

dukess:
Done!

fredlcore:
Perfect, thanks!

PR #131, opened 2020-08-02
RVS61.843E new commands in  range "Ein-/Ausgangstest"

LucekDev:


fredlcore:
Thanks, just a few things:
- DEV_ALL lines should come first, then the device specific lines (as with 7732/7734 etc.).
- Why do you remove DEV_170_ALL for parameter 7805?
- Why do you use VT_CUSTOM_ENUM on 7844 and 7858?
- For clashes (i.e. same command ID for different machines in different parameters), these are only dealt with if the resulting parameter is really irritating. In that case, the conflicting line is given the flag FL_NO_CMD
- For 7717, can you try to set this to a different percentage, like 20 or 40 and log the command? This way, I would assume that only 20% steps are possible, with 0 = 0%, 1=20% and 5=100%.

Actually, I stop here because it takes too much time to go through all the changes. I know that the structure of cmdtbl and the flags and data types is not really intuitive, so maybe it's better if you just send me your serial monitor logs and the parameter plain text options as explained here:
https://1coderookie.github.io/BSB-LPB-LAN_EN/chap10.html

fredlcore:
Two more things: Do not redefine or move around existing defines, like with 7999 by 7844. Other translations will then have to do the same. Just do something like this
`#define ENUM7844_01_TEXT ENUM7999_01_TEXT`
and leave ENUM7999... untouched.
Another thing is the naming of parameters: They are most of the time not 100% exactly the same, but if they are more or less the same, we do not add a new text description or even a new entry in cmdtbl. 
So 
```
#define STR7711_TEXT "Spannungssignal UX1"
#define STR7711_2_TEXT "Ausgangssignal UX1"
```
are basically the same and there is no need to add a new string for that.

LucekDev:
Next days I will be traveling, so I could do corrections when I am back in a week.
1. Ok
2. It is not removed, it is few lines bellow. I inserted block with new commands and in this case not rearranged the line when I was ready.
3. It is VT_CUSTOM_ENUM, VT_ENUM does not work - at the beginning I assumed that 7844, 7858, 7999 will have the same ENUM. For this reason I redefined items of 7999.
4. The same Cmd ID with different parameter number was hiding my parameter, I will try FL_NO_CMD.
5. I am writing from memory, so there are two similar parameters for one I get with VT_PERCENT 5% instead of 100% and second is 3 bytes long - for this I found type which gives proper value, but without %. According specification both are % PWM.

Wiadomość napisana przez fredlcore <notifications@github.com> w dniu 02.08.2020, o godz. 20:05:

﻿

Thanks, just a few things:

  *   DEV_ALL lines should come first, then the device specific lines (as with 7732/7734 etc.).
  *   Why do you remove DEV_170_ALL for parameter 7805?
  *   Why do you use VT_CUSTOM_ENUM on 7844 and 7858?
  *   For clashes (i.e. same command ID for different machines in different parameters), these are only dealt with if the resulting parameter is really irritating. In that case, the conflicting line is given the flag FL_NO_CMD
  *   For 7717, can you try to set this to a different percentage, like 20 or 40 and log the command? This way, I would assume that only 20% steps are possible, with 0 = 0%, 1=20% and 5=100%.

Actually, I stop here because it takes too much time to go through all the changes. I know that the structure of cmdtbl and the flags and data types is not really intuitive, so maybe it's better if you just send me your serial monitor logs and the parameter plain text options as explained here:
https://1coderookie.github.io/BSB-LPB-LAN_EN/chap10.html

fredlcore:
Ok, no worries and no hurry!
As for VT_CUSTOM_ENUM, this was made for telegrams where two parameters are transmitted under one command ID. Can you post a serial log from one of the parameters where you have used VT_CUSTOM_ENUM? Because the way you have defined the ENUM for 7844, it should work with "normal" ENUM. CUSTOM_ENUM addresses a specific byte in the response telegram where more than one kind of information is encoded. This specific byte is contained in the first byte of the ENUM. Since this is zero in your case, it does not differ from a "normal" ENUM and should therefore be used. If it does not work, there are other reasons for that which we would need to address...

fredlcore:
Perfect, thanks!

fredlcore:
Sorry, I accepted the wrong pull request on my mobile phone, so this one is still open...

PR #134, opened 2020-08-04
Category shifting for languages

dukess:
Text enriching if it possible


fredlcore:
Oh, I forgot about these, thanks, very observant!

PR #135, opened 2020-08-04
JSON bugfix: last six caregories not shown

dukess:
JSON bugfix: number of categories was hardcoded as 42. Now is set as CAT_UNKNOWN.


fredlcore:
Thanks!

PR #147, opened 2020-08-17
Working on errors

dukess:
I forgot to add handlers for new data types
Type of prognr 2490 was corrected.


BSB_lan_defs.h
```diff
@@ -8604,19 +8604,19 @@ PROGMEM_LATE const cmd_t cmdtbl2[]={
 {0x053D0821,  CAT_IOTEST,           VT_VOLTAGE,       7711,  STR7711,  0,                    NULL,         FL_RONLY,     DEV_ALL}, // Spannungssignal UX
 
 // 100.0 % PWM
-// HEIZ->DISP ANS 7711 Ein-/Ausgangstest - Ausgangssignal UX1: 00 03 E8 
+// HEIZ->DISP ANS 7711 Ein-/Ausgangstest - Ausgangssignal UX1: 00 03 E8
```
fredlcore, line 8607: Could you check how it is done that this telegram can contain both a frequency and a percentage? The same thing occurs with 7781 where used a percentage type because so far I have never seen the other options (frequency etc.). But if you can access several types (frequency, percentage etc.) with 7711, then the same should be applied to 7781, too.

dukess: Unfortunately, I can't access these parameters via QAA70 even in OEM mode. BSB_Lan shows zeros for these program numbers.

PR #152, opened 2020-09-04
/JA command

dukess:
/JA command added (print 24h average values in JSON)
Bugfix in deprecated function: Writing outside the array


fredlcore:
Thanks!

PR #154, opened 2020-09-19
Preparing for new revolution 

dukess:
* definitions was replaced by variables;
* 'buffer' was removed;
* Virtual programs added: some durations, DHT22 and Dallas sensors can be queried as usual programs (20000-20005, 20100-20299);
* exclude_GPIO[] array was replaced by protected_GPIO[] bit array;
* /JS can set any (not numeric only) parameters now;
* /JI 'dht_bus' value was changed to array of values.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -395,7 +395,7 @@ UserDefinedEEP<> EEPROM; // default Adresse 0x50 (80)
 #endif
 //#include <util/crc16.h>
 #include "src/Time/TimeLib.h"
-#ifdef MQTTBrokerIP
+#ifdef MQTT
 #include "src/PubSubClient/src/PubSubClient.h"
 #endif
 #include "html_strings.h"
@@ -413,10 +413,6 @@ WiFiEspServer *telnetServer;
 #else
 EthernetServer *server;
 EthernetServer *telnetServer;
-#endif
-
-#ifdef MQTTBrokerIP
-IPAddress MQTTBroker(MQTTBrokerIP);
 #endif
 
 Stream* SerialOutput;
@@ -470,7 +466,7 @@ WiFiEspClient max_cul;
 EthernetClient max_cul;
 #endif
 
-#ifdef MQTTBrokerIP
+#ifdef MQTT
 PubSubClient MQTTClient(client);
 #endif
 boolean haveTelnetClient = false;
@@ -872,8 +868,8 @@ void checkSockStatus()
  #endif
    va_end(args);
    switch(debug_mode){
-     case 1: Serial.print(DebugBuff); break;
-     case 2: telnetClient.print(DebugBuff); break;
+     case 1: SerialOutput->print(DebugBuff); break;
+     case 2: if(haveTelnetClient)telnetClient.print(DebugBuff); break;
    }
    return len;
  }
@@ -885,15 +881,15 @@ void checkSockStatus()
    strcpy(DebugBuff, PSTR("\n"));
  #endif
  switch(debug_mode){
-   case 1: Serial.print(DebugBuff); break;
-   case 2: telnetClient.print(DebugBuff); break;
+   case 1: SerialOutput->print(DebugBuff); break;
+   case 2: if(haveTelnetClient)telnetClient.print(DebugBuff); break;
  }
  }
 
  void printToDebug(char *format){
    switch(debug_mode){
-     case 1: Serial.print(format); break;
-     case 2: telnetClient.print(format); break;
+     case 1: SerialOutput->print(format); break;
+     case 2: if(haveTelnetClient)telnetClient.print(format); break;
    }
    return;
  }
@@ -902,13 +898,13 @@ void checkSockStatus()
    #if defined(__AVR__)
    strcpy_P(DebugBuff, format);
    switch(debug_mode){
-     case 1: Serial.print(DebugBuff); break;
-     case 2: telnetClient.print(DebugBuff); break;
+     case 1: SerialOutput->print(DebugBuff); break;
+     case 2: if(haveTelnetClient)telnetClient.print(DebugBuff); break;
    }
    #else
    switch(debug_mode){
-     case 1: Serial.print(format); break;
-     case 2: telnetClient.print(format); break;
+     case 1: SerialOutput->print(format); break;
+     case 2: if(haveTelnetClient)telnetClient.print(format); break;
    }
    #endif
    return;
@@ -918,8 +914,8 @@ void checkSockStatus()
  void printToDebug(uint_farptr_t src){
    strcpy_PF(DebugBuff, src);
    switch(debug_mode){
-     case 1: Serial.print(DebugBuff); break;
-     case 2: telnetClient.print(DebugBuff); break;
+     case 1: SerialOutput->print(DebugBuff); break;
+     case 2: if(haveTelnetClient)telnetClient.print(DebugBuff); break;
    }
    return;
  }
@@ -3123,13 +3119,14 @@ char *GetDateTime(char date[]){
  * Global resources used:
  *
  * *************************************************************** */
+#ifdef LOGGER
 void printTrailToFile(File *dataFile){
  char fileBuf[64];
  // get current time from heating system
  sprintf_P(fileBuf, PSTR("%lu;%s;"), millis(), GetDateTime(date));
  dataFile->print(fileBuf);
 }
-
+#endif
 /** *****************************************************************
  *  Function:  LogTelegram()
  *  Does:      Logs the telegram content in hex to the SD card,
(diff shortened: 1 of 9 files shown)


fredlcore:
Wow, that's a lot of work, thanks! Two questions and one comment:
- Will the code compile of WEBCONFIG is not defined? Because if it is not defined, all these variables will not be declared and I would assume that this will break configuration, wouldnt't it?
- Just out of curiousity: What is the difference between `bus.` and `bus->`?

As for merging this pull request, I will do so once it is generally running (I don't assume it to be bugfree at once ;) ). Before the merge, I will draft a release based on the current, stable codebase and will let people know that "master" will be kind of alpha/beta status for a while.

Thanks again for also lots of other "small" improvements which really make the code more readable and most likely more memory efficient and stable!

dukess:
>Will the code compile of WEBCONFIG is not defined? Because if it is not defined, all these variables will not be declared and I would assume that this will break configuration, wouldnt't it?

I have commented out this part of the code so far, as we need to discuss further actions. However, almost all new variables are used in the main code, so there shouldn't be any problems.

>Just out of curiousity: What is the difference between bus. and bus->?

In the end, there is not much difference. I just redid it before I figured out the bus initialization sequence. In the old version, I would have had to add a few lines.

>As for merging this pull request, I will do so once it is generally running (I don't assume it to be bugfree at once ;) ). 

Today i ran it, so it works for first look :) But i have not DHT22/MAX!/Dallas sensors, so i can't test their functions.

>Before the merge, I will draft a release based on the current, stable codebase and will let people know that "master" will be kind of alpha/beta status for a while.

Sure!

>Thanks again 

You're welcome!

Questions: 
* DHT22 now return string with two values: temperathure and humidity.  What do you think about dividing output into two functions: for example, even for temperature, odd for humidity?
* MAX! sensors AFAIK can return three values, so i can propose split it for three functions.
* I changed the logic to poll One wire bus: command to request new values is now send once in  ONE_WIRE_REQUESTS_PERIOD ms (15 sec) instead command sending every time when temperature is request.
Does it make sense to poll sensors more often than once a minute?

fredlcore:
I'm a bit confused right now, I thought I had answered your questions above already, but now they're not here - did I inadvertedly send them to you via e-mail or was I just daydreaming ;)? In the latter case, I'll reply here again.

dukess:
No, I don't remember what answers were, so it will be better if you repeat it. :)
P.S. Last two commits haven't been tested yet.

fredlcore:
Ok, sorry, maybe I started replying and never finished it ;)...

> Questions:
>     * DHT22 now return string with two values: temperathure and humidity.  What do you think about dividing output into two functions: for example, even for temperature, odd for humidity?

Sounds good to me, especially when they can now be adressed regularly through parameter numbers.

>     * MAX! sensors AFAIK can return three values, so i can propose split it for three functions.

Yes, MAX! sensors return three possible value (types): Current temperature, Destination temperature (both degrees Celsius) and valve opening (in percent).

>     * I changed the logic to poll One wire bus: command to request new values is now send once in  ONE_WIRE_REQUESTS_PERIOD ms (15 sec) instead command sending every time when temperature is request.
>       Does it make sense to poll sensors more often than once a minute?

Hm, there are people who poll parameters every 30 seconds, but I doubt that with temperature values it makes sense to poll these that often. So I would say once every minute is enough. Or does anybody of the other contributors here disagree?

dukess:
Today I ran tests for Due EEPROM and was satisfied with the result: read/write work. So I can ask you to start beta testing.

fredlcore:
I'm currently on vacation until end of this week, so I can't really run tests on site (which would be necessary because I don't want to be the house without controlled heating in case something breaks ;) ). 
I think I'll make a release with the current stable version and then merge your PRs into Master so that others can test, too. What do you think?

dukess:
Have a nice vacation!

This way suits me perfectly.

dukess:
Just FYI: i create new action/workflow file for automatic language (.js, .js.gz, .json) files regeneration.

fredlcore:
Hi @dukess, because of the changes I made, there is now a conflict so that I can't merge automatically.
Also one question regarding the MAC address in BSB_lan_config.h: This seems to have gone there - at least on my shield it's all zeros if the variable is not set, and this could result in improper network configuration. Should we set the current default until its changed in the webinterface (if one can change it there, I don't know)?

As for the manual: Coud you give @1coderookie a few pointers where current users **have** to make a change in order to get the new version running?
Also it would be immensely helpful if we could get a list of new configuration options _in the _config.h_ that would need to be explained in the manual. 
Anything that can be configured in the webinterface can be explained at a later stage because most people will somehow get that, I think...

dukess:
Hi. When switching to a new branch, you will have to use a new configuration file that is incompatible with the old one.
Default MAC defined as 
`byte mac[] = { 0x00, 0x80, 0x41, 0x19, 0x69, 0x90 };`
in new config.h

fredlcore:
Is the new file named "config.h" or is the name "BSB_lan_config.h.default" still accurate? I couldn't find a file named "config.h" in the pull request, that's why I ask...

fredlcore:
Ok, forget about it, I found it!

dukess:
:)

fredlcore:
Ok, once the remaining conflicts are resolved, I'll merge - really excited :)!

dukess:
Ready.

fredlcore:
Done!

dukess:
Look at the changes, please, as soon as possible. We need to discuss ways to move forward.

fredlcore:
Compiling right now, here are the warnings that are thrown:
```
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:443:0: warning: "OUTBUF_USEFUL_LEN" redefined
 #define OUTBUF_USEFUL_LEN (OUTBUF_LEN * 2)
 
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:441:0: note: this is the location of the previous definition
 #define OUTBUF_USEFUL_LEN (OUTBUF_LEN)
```

```
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino: In function 'void implementConfig()':
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:3416:36: warning: format '%x' expects argument of type 'unsigned int*', but argument 3 has type 'char*' [-Wformat=]
       sscanf(outBuf + i, "%02x", &c);
                                  ~~^
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino: In function 'void queryVirtualPrognr(int, int)':
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:4799:15: warning: unused variable 'log_sensor' [-Wunused-variable]
        size_t log_sensor = tempLine / 4;
               ^~~~~~~~~~
```

fredlcore:
Hm, my Due does not respond, IP is not accessible (I'm using fixed IP), and no output on serial monitor...

fredlcore:
Ok, upon a reset, it gets up to the point where it has finished calculating the free space on SD card...

fredlcore:
I've added debug markers, and it seems to stop after `server->begin;` in the setup() function. Strangely, it outputs the marker, but not the line feed...

fredlcore:
I have now tried with DHCP instead of fixed IP and now it works. This was my previous configuration:
```
uint16_t HTTPPort = 80;
boolean useDHCP = false;
byte ip_addr[4] = {192,168,1,50}; // please note the commas instead of dots!!!  Set useDHCP to true when you want DHCP
byte gateway_addr[4] = {0,168,1,1}; // this is usually your router's IP address. Please note the commas instead of dots!!! Ignored, when first value is 0
byte dns_addr[4] = {0,168,1,1}; //  DNS server. Please note the commas instead of dots!!! Ignored, when first value is 0
byte subnet_addr[4] = {0,255,255,0}; // Please use commas instead of dots!!! Ignored, when first value is 0
```
Once I set useDHCP to true it works, but now I don't know which IP address was assigned. But other than that, fixed IP assignment should also work ;)...

dukess:
On the one hand, we can create an Ethernet object without a subnet and gateway (with further configuration), on the other hand, I did not test the statics.
Need to read more manuals.

fredlcore:
I now found out the DHCP-assigned IP address but I'm not sure if what I see in the Configuration menu is correct:
Bildschirmfoto 2020-11-10 um 22 24 31
At the bottom, I have a submit button with one editable option which is Reading from EEPROM on/off. That one value is stored correctly. Should there be more options? No stress, just to make sure that I'm not missing something.

There also seems to be a glitch with:
```
Loggen der folgenden Parameter alle 3600
Store parameters on disk: Ja
Sekunden: 
```
It basically says 
"Storing the following parameters every 3600
Store parameters on disk: Yes
seconds"

So the "seconds" should come right after the 3600 and the "Store parameters on disk: Ja" should probably be a different line or something?

Other than that: Really, really great work! Just browsing through the code gives me an impression of how much work you put in there and how much more sense things make now. It'll take time for me to get a complete understanding of all the new hooks, but it looks really great! For now, I'll concentrate on bugfixes and getting the install up and running because that will be what people will bug me with, especially now in winter when everybody starts tweaking their heating systems ;)...

fredlcore:
Oh, and one more minor bug: The version is now 2.0.0, but the version check on the main page still alerts me that there is a newer version 1.1.51 out there...

fredlcore:
Ok, I did some testing:
It really is the line of 
`Ethernet.begin(mac, ip);`
which lets
`server->begin;`
come to a halt. Once just call 
`Ethernet.begin(mac);`
it works fine. The weird thing is that even if I copypaste the old network code in there or even hard-code my ip address, it still comes to a halt. But you didn't update the Ethernet library in any way, so I wonder why this is the case.

fredlcore:
Another error: Compilation fails if LOGGER and WEBSERVER are disabled:
```
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:700:19: error: variable or field 'transmitFile' declared void
   return EEPROMwasChanged;
                   ^~~~
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:700:19: error: 'File' was not declared in this scope
/Users/frederik/Documents/PlatformIO/Projects/BSB-LAN/src/BSB_lan.ino:700:19: note: suggested alternative: 'Wire'
   return EEPROMwasChanged;
                   ^~~~
                   Wire
```
I was thinking that maybe the Ethernet problems may have something to do with the SD card functionality and therefore wanted to disable it, but then this error comes.

fredlcore:
Switiching back to Ethernet Library version 1.1.2 (plus removing a few lines of incompatible code for setting gateway etc.) partially works with fixed IP, at least to the point that it doesn't halt at "server->begin;" anymore. However, the IP address is not assigned correctly, it seems, as Ethernet.localIP() outputs "255.255.255.255" (or my router is now confused).
So the question is why the library all of a sudden stops working with fixed IPs if it has been working fine until the update?

fredlcore:
Another minor strange observation:
`SerialOutput->print(F(" MB free\n"));`
prints a lot of empty spaces after the "free". When removing the `\n` and replacing `print`with `println`, the output is as expected and outputs a newline right after the "free". Not sure if this has any other implications.

dukess:
I prepare patch now, tested it with static IP - it works. But i catch big trouble with Mega: Assembler shows many messages like this
C:\Users\User\AppData\Local\Temp\ccI8BaTZ.s:63977: Error: value of 71409 too large for field of 2 bytes at 41050
C:\Users\User\AppData\Local\Temp\ccI8BaTZ.s:68605: Error: value of 71409 too large for field of 2 bytes at 47102

Now i debug it.

SerialOutput->print(F(" MB free\n"));
for this we need add \r symbol:
SerialOutput->print(F(" MB free\n\r"));
or call println, sure.

fredlcore:
Ok, great, thanks! Did you have to patch the Ethernet library for that or was it code in BSB_lan.ino? In case patching the Mega errors takes longer, could you give me a pointer where to make changes because I only have one system here and my heating is uncontrolled since yesterday evening now ;) (I do have a Mega here for previous test runs, but now I want to try out the new Due features of course, so it's woking on the "open heart" for me now ;) ...)

dukess:
Code in BSB_lan.ino. 
On Due code compiled and run without hard troubles (except fresh bugs), but you can rollback to v1.1. New code can be compiled for Mega with disabled WEBCONFIG option. All glithches with assembler code was introduced in this module.

fredlcore:
Thanks!

PR #158, opened 2020-10-08
Setup basic Github Actions workflow

fl42:
Hello,

This PR sets up a basic Github Actions workflow to ensure the code in the repo can be built.

### DIFF
--- a/.github/workflows/platformio.yaml
+++ b/.github/workflows/platformio.yaml
@@ -1,0 +1,48 @@
+name: PlatformIO CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python_version:
+        - "3.6"
+        - "3.7"
+        - "3.8"
+
+    steps:
+    - name: Checkout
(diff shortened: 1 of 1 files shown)


fredlcore:
I'm not sure what you mean with "to ensure the code in the repo can be built". It buildsfine with both the Arduino IDE as well as PlatformIO/VScode. What does this file do and why is it necessary to be added?

fl42:
Hi,

Building code on each commit or PR using Github Actions ensures that the code can be built repeatedly (i.e. on anyone's computer).
This is one of [Continuous Integration](https://en.wikipedia.org/wiki/Continuous_integration#Compile_code_in_CI) best practices.

This PR sets up a really basic CI (ensure the software can be compiled), in the future we may add another steps: static code analysis, various tests, etc.
For instance, when a PR is open, you will have a flag telling you if the new code breaks the compilation.

You might be interested to see how other projects take advantage of CI (e.g. [ESPHome](https://github.com/esphome/esphome/blob/dev/.github/workflows/ci.yml)).

fredlcore:
Ah, ok, thanks, I have not really worked with this before, but have merged your PR now.

PR #159, opened 2020-10-09
fast fix for https://github.com/fredlcore/bsb_lan/issues/157

dukess:
Query for 20000-20200 still break MQTT i mean


fredlcore:
Great, thanks a lot!

PR #161, opened 2020-10-18
MQTTPayload exists only if MQTT_JSON is defined

fl42:
Fix #157 
@dukess Any opinion on this fix?

Issue #157, opened 2020-10-08
Bad JSON in MQTT message


fl42:
Hi,

I tried to publish some values using MQTT with JSON.

Unfortunately the published JSON looks invalid:
```json
{"BSB-LAN":{"status":{"8700":"17.6","8830":"51.9","8310":"57.8","8744":"---","8950":"57.8","8331":"14568",}}
```
(note the missing '}')

My broker is mosquitto version 1.6.12
I'm using BSB Lan version 1.1.39-20201005105852 on Arduino Due.

My config (logging part):
```h
/* activate logging on SD-card. Requires a FAT32-formatted Micro-SD card inserted into the Ethernet-Shield's card slot */
// #define LOGGER
// Logging von "rohen" Bus-Datentelegrammen
// Telegrams logging. False - disabled on start, true - enabled on start.
#define LOGTELEGRAM false
// You can enable logging with URL /LE command or disable it with /LD command when run time.
// Data will be write to "journal.txt"
boolean log_unknown_only = 1;         // should we log only unknown commands when logging bus telegrams?
boolean log_bc_only = 0;              // should we log only broadcast commands (dest = 0x7f) when logging bus telegrams?


// Logging data from parameters
// Interval and list of parameters can be redefined through /L command when run time
// Data will be write to "datalog.txt"
int log_parameters[20] = {
  8700,
  8830,
  8310,
  8744,
  8950,
  8331
};
unsigned long log_interval = 30;    // logging interval in seconds

// Activate sending log_parameters to MQTT broker every log_interval seconds
#define MQTTBrokerIP 192,168,2,1	// Please use commas instead of dots!!!
//#define MQTTUsername "User" // Set username for MQTT broker here or comment out if no username/password is used.
//#define MQTTPassword "Pass" // Set password for MQTT broker here or comment out if no password is used.
#define MQTTTopicPrefix "bsb_lan" 	// Optional: Choose the "topic" for MQTT messages here
#define MQTT_JSON 					// Optional: Use this if you want a json package of your logging information printed to the mqtt topic
//#define MQTTDeviceID "MyHeater"	// Optional: Define a device name to use as header in json payload. If not defined, BSB-LAN will be used.
// Payload will be of the structure: {"MQTTDeviceID": {"status":{"log_param1":"value1","log_param2":"value2"}, ...}}
```

Have any of you ever had the problem?
How can I help identify the root cause?

Many thanks

fredlcore:
[...]

PR #162, opened 2020-10-18
[CI] Try to compile with MQTT and MQTT JSON

fl42:
To spot a bug like #157 before a user notices it :)
CI is expected to fail until #161 is merged

### DIFF
--- a/.github/workflows/platformio.yaml
+++ b/.github/workflows/platformio.yaml
@@ -12,6 +12,10 @@ jobs:
         - "3.6"
         - "3.7"
         - "3.8"
+        bsb_lan_config:
+        - BSB_lan_config_default.h
+        - BSB_lan_config_mqtt.h
+        - BSB_lan_config_mqtt_json.h
 
     steps:
     - name: Checkout
@@ -41,8 +45,7 @@ jobs:
 
     - name: Setup config
       run: |
-        # Use default config
-        cp BSB_lan_config.h.default BSB_lan_config.h
+        cp tests/config/${{ matrix.bsb_lan_config }} BSB_lan_config.h
 
     - name: Run PlatformIO CI
       run: platformio ci --board=due .
--- a/tests/config/BSB_lan_config_default.h
+++ b/tests/config/BSB_lan_config_default.h
@@ -1,0 +1,1 @@
+../../BSB_lan_config.h.default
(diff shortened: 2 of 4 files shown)


fredlcore:
I still have to fully understand the whole GitHub actions subject, but this looks reasonable to me. I have merged #161 now, but the checks here still fail. How can I run them again?

fl42:
I just rebased my fork to your master branch, should work now

fredlcore:
Yes, great, thanks!

fredlcore:
I'm not sure if this is related to your fix, but I got a user who reported that when renaming BSB_lan_config.h.default to BSB_lan_config.h (instead of copying it to a new file and leaving BSB_lan_config.h.default untouched), the compiler breaks with an error. 
I think it should be possible to just rename the file (because that's what it says in the manual), but if that would render the tests useless, then we should change the manual (cc'ing @1coderookie here, just in case).

fl42:
Hi @fredlcore,

Should not be related to the tests as I did not change anything to the code or how the config is managed.
On my side I'm able to compile using plateformio without BSB_lan_config.h.default.

The best would be to open an issue with the error message so we could investigate @1coderookie?

fredlcore:
Not sure if I can get him to do that so quickly, but here's what he wrote in German:

Ein kleines Problem hatte ich. Die BSB_lan_config.h.default hatte ich entsprechend der Anleitung umbenannt. Damit lässt sich das Projekt aber nicht kompilieren, da der Link tests/config/BSB_lan_config_default.h dann auf die nicht mehr existierende Datei BSB_lan_config.h.default verweist. 

Since it mentions the tests folder,I assume it has nothing to do with the main code.
Even though most of us are not using the Arduino IDE, this is what 95% of ordinary users will be using, so we need to make sure it compiles there without errors...

fl42:
Got it, sorry for the issue
Actually tests/config/BSB_lan_config_default.h is a symbolic link to BSB_lan_config_default.h
If BSB_lan_config_default.h is renamed, Arduino IDE fails because of the broken symlink (even if it does not need this symlink)
I'll open a PR to fix it, thanks for the head's up

fredlcore:
Thanks!

fredlcore:
Since now there is no symlink anymore, does this mean I have to manually update BSB_lan_config.h in both directories (main and tests)?

fl42:
Yes,
In tests/config/ there are the configurations that will be used (successively) to build the code.

At the moment the CI makes sure that the project build with the default config, the config with MQTT without JSON, and the config with MQTT and JSON.

PR #171, opened 2020-11-11
Bugfixes after revolution

dukess:
Bug fixes:
Board can't start with static IP now. Fixed.
UseEEPROM moved to config file. UseEEPROM = 0 can be used for fallback startup. Do not forget update config file.
Wrong info about logging and interval on config page.

Known issues:
Mega 2560 board: if program size more than ~245000 bytes you can got error "avrdude: verification error, first mismatch at byte 0x3c000"

WEBCONFIG module doesn't  work at this time on Mega:
C:\Users\User\AppData\Local\Temp\ccuOzsoN.s:115706: Error: value of 72118 too large for field of 2 bytes at 112281
C:\Users\User\AppData\Local\Temp\ccuOzsoN.s:115849: Error: value of 72118 too large for field of 2 bytes at 112468

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -584,8 +584,6 @@ uint16_t pps_values[PPS_ANZ] = { 0 };
 boolean time_set = false;
 uint8_t current_switchday = 0;
 
-byte UseEEPROM = 0;
-
 #include "BSB_lan_EEPROMconfig.h"
 
 static int baseConfigAddrInEEPROM = 0; //offset from start address
@@ -690,8 +688,7 @@ boolean writeToEEPROM(uint8_t id){
   if (!EEPROM_ready) return false;
   boolean EEPROMwasChanged = false;
   if(!options[id].option_address) return false;
-  uint16_t len = getVariableSize(id);
-  for(uint16_t i = 0; i < len; i++){
+  for(uint16_t i = 0; i < getVariableSize(id); i++){
     if(options[id].option_address[i] != EEPROM.read(i + options[id].eeprom_address)){
       EEPROM.write(i + options[id].eeprom_address, options[id].option_address[i]);
       EEPROMwasChanged = true;
@@ -702,15 +699,13 @@ boolean writeToEEPROM(uint8_t id){
 
 boolean writeToConfigVariable(uint8_t id, byte *ptr){
   if(!options[id].option_address) return false;
-  uint16_t len = getVariableSize(id);
-  memcpy(options[id].option_address, ptr, len);
+  memcpy(options[id].option_address, ptr, getVariableSize(id));
   return true;
 }
 
 boolean readFromConfigVariable(uint8_t id, byte *ptr){
   if(!options[id].option_address) return false;
-  uint16_t len = getVariableSize(id);
-  memcpy(ptr, options[id].option_address, len);
+  memcpy(ptr, options[id].option_address, getVariableSize(id));
   return true;
 }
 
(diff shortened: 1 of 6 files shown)


dukess:
Do not apply  now, please

fredlcore:
Thans for the hint with the first byte mismatch error! I had this a few weeks back and thought that my Mega was broken, but if this is a "normal" error, then I know I can keep it :)...
Will wait for the merge until I get your "go ahead"...

dukess:
This patch do something worse: now i can't get access through web and telnet. I need to take a walk and clear my head.

fredlcore:
Ok, take your time!

fredlcore:
Not sure if that helps, but it seems to be a certain time until the Arduino hangs: When I add output to the Serial port like this:
```
  Serial.println(1);
  Serial.println(2);
  server->begin();
  Serial.println(3);
  Serial.println(4);
```
It prints the "1" and _sometimes_ the "2", but that's it. Even when I put the number output further down in the code. Once it gets to the 1-Wire initialization, it's over and not even the "1" gets printed anymore. Very strange.

fredlcore:
I have tried copying the function calls in Ethernet.begin from the (mac) to the (mac, ip, dns, gateway, subnet) function, and if I do a 1:1 copy, it works, but if I set the fixed IP hard-coded into the function, it somehow always gives me back 255.255.255.255. But it works. Maybe it would be worthwhile to investigate further what the difference is with the DHCP and the fixed IP begin()-functions?

dukess:
Once idea i have: one (or more variables) will overwrite data of other objects.
Now i try build and run source from different times.

Revolution is very bloody.

dukess:
I think to write small sketch with network testing because last more or less stable version was version before moving parameters from definitions to variables.

fredlcore:
> Revolution is very bloody.

:)

dukess:
I want to try one more time.

fredlcore:
Means I should merge or not (yet)?

dukess:
Short manual:
Add byte UseEEPROM = 1; in config
Enable 
#define CONFIG_IN_EEPROM
#define WEBCONFIG

Flash program

Start.
Call /C program, change config, save changes.
For store settings in EEPROM call /CS

dukess:
You can merge

fredlcore:
Great, fixed IP works, wonderful!
Did you find out what the problem was?

dukess:
First problem: declared too small array for options table.
Second: deleted non-exist MQTTClient object (by fact - erased RAM in random addreses).

The main reason is an excessive passion for optimization %)

fredlcore:
Ok :)...
I found a few more bugs, unfortunately for PPS, so we both will have to do the debugging:
When running in PPS mode, the on and off times have been read from EEPROM before already. Now I get strange time slots , they're all 42:30 hours, which means that a value of 0xFF is being read from EEPROM. It should somehow be made sure that this part of the EEPROM is zeroed out upon first start (or maybe easier: whenever there is a value higher than 144/0x90.
Also, the value for the QAA model to be sent for PPS telegram type 0x38 is not taken from the variable QAA_TYPE (which would be either 0x53 or 0x52), but it's also 0xFF.
I'll try to fix the latter now because otherwise PPS won't work at all because the heater needs to see the QAA type value before it sends or accepts data...

fredlcore:
Ok, everything's fine once the EEPROM is overwritten with zeroes :). So a /JE should be issued on a fresh board. Do you think it would be feasible to check if the EEPROM has ever been zeroed-out? For example by looking for 0xFF in various places and then zero out the EEPROM?

fredlcore:
...and the web-configuration is awesome!!!

dukess:
Thank you for testing!

The problem with the bus is that I read the entire array. But, as I understand it, we need to read only part of the elements, the rest are used in the current work and should not be saved between sessions.

PR #174, opened 2020-11-14
\r\n everywhere. Room device type selector added.

dukess:
\n replaced by \r\n everywhere.

Room device type selector added.
pps_settings is now not fully saved, only parameters from PPS_TWS to PPS_BRS.
After merging this patch you will need to reconfigure device bacause i change schema version not. CRC of current schema will changed and EEPROM config not readed.


fredlcore:
Thanks for the update - I'm thinking about the room type selector. Should we hard code the parameter number here? In case we have to move around PPS parameters (not very likely, but there are still a few unknown telegrams which may have to be sorted in before 15062) we have to remember that this hard coded parameter needs to be changed. 
On the other hand, we know that so far there are only two states that this variable can have (maybe a third one), so should we rather hard code these parameters here (or elsewhere)?

dukess:
what about 
`               i=findLine(15000 + PPS_QTP,0,NULL); //return ENUM15062 (device type)`
instead
`               i=findLine(15062,0,NULL); //return ENUM15062 (device type)`?

In future we probably can got broken configuration (and boot with defaults from config file) after one of next update.
We have 68 PPS commands now. I can propose expand pps_values array up to 128 values and forget it for many years. :-)

fredlcore:
Two great ideas :)!

fredlcore:
The thing about the commands is however the order, and not the total sum. For example, if we find another temperature parameter, it may have been sorted into index 20++ or so. But for that the "15000 + PPS_QTP" is great!

dukess:
One problem stills here:  strings ID will be changed when programs will be reorder. I think here no common decision.
Unless only the strings are called differently like STRPPS_QTP.

fredlcore:
Yes, but that's something that I always have to keep in mind (and which you see once you add the text). But to remember in a few months that there is this one number in the code that I have to change, that is more difficult. So this solution is great now, thanks!

PR #195, opened 2020-11-29
Enhanced software-controlled writing values to programs

dukess:
Now is three states here:
0 - all programs is read-only.
1 - all programs is writable except read-only and OEM
2 - all programs is writable except read-only

WARNING: "read-only" behavior is inverted after upgrade. Check setting.

### DIFF
--- a/BSB_lan.ino
+++ b/BSB_lan.ino
@@ -417,7 +417,7 @@ UserDefinedEEP<> EEPROM; // default Adresse 0x50 (80)
 #endif
 
 boolean EEPROM_ready = true;
-byte readOnlyMode = 1;
+byte programWriteMode = 0; //0 - read only, 1 - write ordinary programs, 2 - write ordinary + OEM programs
 
 #ifdef WIFI
 WiFiEspServer *server;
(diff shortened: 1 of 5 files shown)


fredlcore:
What exactly is inverted?

fredlcore:
Ok, got it from the source code - one question: Does this mean you can enable OEM writing just via the webinterface?
I think it's ok to be able to change this in _config.h, but I would not enable to change this to OEM in the webinterface, so that people don't accidentially enable this and later complain. If they change it in the _config.h, it's more or less the same as before, so that would be fine.

dukess:
If  readOnlyMode == 1 (all parameters is read-only) on previous installation then now 
it switch to mode "all programs is writable except read-only and OEM".
If  readOnlyMode == 0 (all programs is writable except read-only and OEM) on previous installation then now 
it switch to mode "all parameters is read-only".

dukess:
>Does this mean you can enable OEM writing just via the webinterface?
I think it's ok to be able to change this in _config.h, but I would not enable to change this to OEM in the webinterface, so that people don't accidentially enable this and later complain. If they change it in the _config.h, it's more or less the same as before, so that would be fine.

"Yes" and "No". 
User need to set FL_SW_CTL_RONLY as DEFAULT_FLAG to switch access mode through web interface. But if user set it then access mode chan be switchced without warnings.

fredlcore:
Ok, got it!
@1coderookie: Could you adjust the manual accordingly?
I'll post a note in the FHEM forum so that existing users know they need to make a change in their configuration.

dukess:
Thank you!

1coderookie:
Sure! I didn't really understand the 'read only' and 'writeable except read-only' difference, but I'll wait and see what you'll post in the forum ;)

fredlcore:
@dukess: I just flashed the most recent version (haven't flashed for 1-2 weeks) and I had changed my BSB-LAN board in the meantime, meaning there was an empty (all 0xFF) EEPROM inserted. Once I had flashed the Arduino and boot it, it hangs at "Reading EEPROM...". I then changed useEEPROM in _config.h to 0, then it booted fine. I then did a /NE to write the flash with all zeros, set useEEPROM back to 1 and then it all was fine. Could it be that an empty (all 0xFF) EEPROM is creating troubles? If so, we should think about checking if the EEPROM is empty (all 0xFF) and then clear it with all zeros.

dukess:
I had same trouble some time ago and add some debug outputs and checks so this problem is disappear.
Not ready to do answer instantly, need to make some tests.

dukess:
@1coderookie 
"0 - all programs is read-only" is equal DEFAULT_FLAG FL_RONLY in config file.
"1 - all programs is writable except read-only and OEM" equal DEFAULT_FLAG 0
"2 - all programs is writable except read-only" equal DEFAULT_FLAG 0 and FL_OEM 4

PR #199, opened 2020-12-02
Bugfix: hanging on new EEPROM

dukess:
Note: I change pattern for EEPROM erasing from 0x00 to 0xFF for testing on "new" EEPROMs and with PPS settings. 

For issue https://github.com/fredlcore/bsb_lan/issues/196


fredlcore:
Ok, but later on we will revert back to 0x00? And shouldn't we then add a clear EEPROM instruction  during setup if version is 0xFF?

PR #200, opened 2020-12-02
init pps_values[PPS_QTP] by QAA_TYPE at start

dukess:


BSB_lan.ino
```diff
@@ -7333,11 +7333,14 @@ uint8_t pps_offset = 0;
                 if (p[4]=='1') { //Use EEPROM
                   UseEEPROM = 0x96;
                 } else {   //Use _config.h
-                  UseEEPROM = 0;
+                  UseEEPROM = 0x69;
```
fredlcore, line 7336: Are you sure this is 0x69 and not 0x96?

dukess: Yes. Two values now: 0x96 will enable and 0x69 will disable EEPROM using. This is for case when EEPROM was erased with "0". If i leave UseEEPROM = 0 then EEPROM will be disabled by default.

fredlcore: Got it, thanks!

PR #206, opened 2020-12-06
Code reduction #2. Option I_WILL_USE_EXTERNAL_INTERFACE  to disable /DG

dukess:
you can add #define I_WILL_USE_EXTERNAL_INTERFACE to config file for saving space (~6..8Kb). /DG command will be disabled

### DIFF
--- a/BSB_lan_config.h.default
+++ b/BSB_lan_config.h.default
@@ -266,6 +266,15 @@ byte destinationServer[128] = ""; //URL string to periodically send values to an
 uint16_t destinationPort = 80; //port number for abovementioned server
 uint32_t destinationDelay = 84600; //seconds interval to send values
 
+//For Arduino Mega users:
+//If You use CONFIG_IN_EEPROM and WEBCONFIG modules then
+//You can enable I_DO_NOT_WANT_URL_CONFIG for saving space (~1.2Kb).
+//This will disable configuration through URL commands (/A, /L, /P)
+//#define I_DO_NOT_WANT_URL_CONFIG
+//If You defined WEBSERVER and use external web interface on SD-card then you can
+//enable I_DO_NOT_WANT_URL_CONFIG for saving space (~6,8Kb). /DG command will be disabled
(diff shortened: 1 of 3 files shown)


BSB_lan_config.h.default
```diff
@@ -266,6 +266,15 @@ byte destinationServer[128] = ""; //URL string to periodically send values to an
 uint16_t destinationPort = 80; //port number for abovementioned server
 uint32_t destinationDelay = 84600; //seconds interval to send values
 
+//For Arduino Mega users:
+//If You use CONFIG_IN_EEPROM and WEBCONFIG modules then
+//You can enable I_DO_NOT_WANT_URL_CONFIG for saving space (~1.2Kb).
+//This will disable configuration through URL commands (/A, /L, /P)
+//#define I_DO_NOT_WANT_URL_CONFIG
+//If You defined WEBSERVER and use external web interface on SD-card then you can
+//enable I_DO_NOT_WANT_URL_CONFIG for saving space (~6,8Kb). /DG command will be disabled
```
fredlcore, line 275: That should say "I_WILL_USE_EXTERNAL_INTERFACE here, shouldn't it?

fredlcore:
Do you _have_ to define WEBSERVER in order to disable /DG? I would think that quite a few users don't need the graph display because they use FHEM or some other home automation, but don't want to use WEBSERVER.

dukess:
User don't need to enable the webserver to delete /DG.

fredlcore <notifications@github.com> 6 декабря 2020 г. 16:31:36 написал:

fredlcore:
Ok, I was wondering because the comment in _config.h.default says that "If you defined WEBSERVER...". So we can remove that condition?

dukess:
Ok. I just assumed that "normal" users wouldn't do it.

fredlcore:
I think there are quite a few Mega users who would really like to use 2.0 version but don't need the graphical display because they can download the logs from the SD card and display them otherwise.

PR #224, opened 2020-12-14
Update BSB_lan.ino

dukess:
Append ENUMs with line value 65535 and description "---".
VT_ENUM will return 65535 when query return "disabled"

SET function not implemented


dukess:
Web output without --- output in list. Need to rewrite HTML building code.

PR #229, opened 2020-12-16
Update BSB_lan.ino

dukess:
set(): implement "disable" for ENUMs.

@fredlcore i concerned with this fragment:
```
      if (val[0] == '-') {
        t=((int)(atof(val)*operand));
      } else {
        t=atoi(val)*operand;
      }
```
if val is positive then atoi() cut fractional part. Can you check it?


fredlcore:
Sure, you're right, it has to be atof, not atoi, my bad...

PR #231, opened 2020-12-19
Update BSB_lan.ino

dukess:


BSB_lan.ino
```diff
@@ -6280,7 +6280,7 @@ uint8_t pps_offset = 0;
             if ((httpflags & HTTP_ETAG))  { //Compare ETag if presented
               if (memcmp(outBuf, outBuf + buffershift, sprintf_P(outBuf + buffershift, PSTR("\"%02d%02d%d%02d%02d%02d%lu\""), dayval, monthval, lastWrtYr, FAT_HOUR(d.lastWriteTime), FAT_MINUTE(d.lastWriteTime), FAT_SECOND(d.lastWriteTime), filesize) + 1)) {
                 // reuse httpflags
-                bitClear(httpflags, HTTP_AUTH+HTTP_GZIP); //ETag not match
+                bitClear(httpflags, 3); //HTTP_ETAG bit. ETag not match
```
fredlcore, line 6283: Would 
`httpflags &= ~HTTP_ETAG;`
work as well?

BSB_lan.ino
```diff
@@ -6261,7 +6261,7 @@ uint8_t pps_offset = 0;
           if ((httpflags & HTTP_GZIP)) dataFile = SD.open(urlString + ".gz");
           if (!dataFile) {
             // reuse httpflags
-            bitClear(httpflags, HTTP_AUTH); //can't use gzip because no gzipped file
+            bitClear(httpflags, 1); //HTTP_GZIP bit. can't use gzip because no gzipped file
```
fredlcore, line 6264: Would
httpflags &= ~HTTP_GZIP;
work as well?

BSB_lan.ino
dukess: Here is bitClear(). In this function we set bit position not value.

BSB_lan.ino
dukess: Same as above.

BSB_lan.ino
fredlcore: Yes, I know, but if I understand your code correctly, you want to clear the 3rd bit in httpflags. HTTP_ETAG is defined as 8. So by ANDing NOT 8 it should have the same effect and we can use a more human readable definement.

dukess: Yes, this will do

dukess:
I implement httpflags &= ~HTTP_GZIP; and httpflags &= ~HTTP_ETAG; in other patch.

PR #232, opened 2020-12-19
Program size optimization.

dukess:
small optimization in set()
Web server optimization. Save up to 1300 bytes flash and 30 bytes RAM.
Fix for IPWE (and, possible, other): when query failed decodedTelegram contain no description, unit name and other fields.
/Q size optimization. ~1 Kb flash and 30 bytes of RAM saved


fredlcore:
Hi @dukess, thanks for all these efforts, and I'm all for optimizing memory, both RAM and Flash, but I think this affects code-readability in several cases a bit too much for the moderate amount of bytes saved. 
When debugging code, it's not really intuitive to read something like this:
`printHTTPheader(200, 6, true, false, -1);`
instead of
`printToWebClient(PSTR("HTTP/1.1 200 OK\r\nContent-Type: application/json; charset=utf-8\r\n\r\n{\r\n"));`

The same goes for cases like this:
```
   print_aHref();
   printToWebClient(PSTR("' ID=main_link>BSB-LAN Web</A></h1></center>\r\n"));
   printToWebClient(PSTR("<table align=center><tr bgcolor=#f0f0f0>"));
   print_header_TD();
```

As a general rule of thumb, I would happily accept optimizations where abstract tasks are delegated to functions, but not so much specific functions that are themselves called in specific functions. In case anyone needs to debug these functions, one has to do it in several places at the same time and that is not really intuitive.

Also, why did you revert my changes here:
```
      if (val[0] == '-') {
        t=((int)(atof(val)*operand));
      } else {
        t=atof(val)*operand;
      }
```
to:
```
      t=atof(val)*operand;
``` 
?

My compiler gives me a warning when trying to convert a negative float to an unsigned int, see here:
https://www.embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/
The result of the multiplication will have to be cast to an int before assigning it to the unsigned int. 

The same goes for this:
`uint32_t t=((int)(atof(val)*operand));`

dukess:
> printHTTPheader(200, 6, true, false, -1);

The possible complexity of readability is compensated by the ease of debugging: the function does not contain complex logic and the result is immediately visible in the browser.

Another reason is that I have a necessity to do compact caching management for different data types: text files (datalog, journal), JSON and internal interface must be without caching; icon and pictures can be cached for long time etc.
Moreover, i would like to experiment with caching part of JSON, more precisely - with the commands /JK=ALL, /JK=<x> and /JC=<x>: despite the fact that they are generated on the fly, their content changes quite rarely.

>print_aHref();
>print_header_TD();

This is my weakness: when I see something like a lot of lines "<td class=\"header\" width=20% align=center>" i lost control

> Also, why did you revert my changes here:

Thank you for explanation. I was sure it would work without any quirks. I will revert back it to your variant.

dukess:
> This is my weakness: when I see something like a lot of lines "<td class="header" width=20% align=center>" i lost control

Can you move  width=20% align=center and some other to CSS?

fredlcore:
Yes, it could be moved to CSS, but wouldn't that cost more space all in all? It should not require a SD card, so we would need a handler etc.?
And agreed with printHTTPheader, but could we use definements or something like that so that the call would be more readable, so instead of
`printHTTPheader(200, 6, true, false, -1);`
we sould have something like:
`printHTTPheader(200, APPLICATION_JSON, ADD_CHARSET, IS_GZIP, -1);`
or similar?

dukess:
>Yes, it could be moved to CSS, but wouldn't that cost more space all in all? It should not require a SD card, so we would need a handler etc.?

No. I look in code. HTML page header already contain CSS section so you can use it.

>printHTTPheader(200, APPLICATION_JSON, ADD_CHARSET, IS_GZIP, -1); 

Sure!

fredlcore:
You are knowing my code better than me sometimes :) - ok, will add this as CSS, but probably not before Christmas/New Years... Will have to add the VT_CUSTOM_BIT field first because a significant new parameter we found will need it.

fredlcore:
Great work, as always, thanks a lot!

PR #240, opened 2020-12-27
Config variables for WiFi added to config schema.

dukess:
WiFi SSID and Password (Schema version 4).


fredlcore:
Thanks! Is the Wifi definement still needef or is it also handled via webconfig?

dukess:
Definement needed as old WiFi.

PR #244, opened 2020-12-29
Update BSB_lan_config.h.default

dukess:


fredlcore:
Thanks, but I think we should keep the configuration reduced to the most relevant information. Due users won't have the need to save RAM for a long time and MEGA users should rather disable the respective features. Maybe this information is better placed in the manual?

dukess:
It would be good decision

dukess:
@1coderookie Ulf, can i ask you to add this text to anhang_d?

```
Also you can reduce size of some variables for freeing RAM or enlarge their size if You need. Ex: PASSKEY[], avg_parameters[], log_parameters[], ipwe_parameters[], max_device_list[].
But be aware: if You change size of variables on configured system your configuration will be lost and default values will be loaded from this file.
You can backup configuration to JSON and restore it. (JSONCONFIG module needed). 
```

1coderookie:
Yepp, will do it asap, thanks!

PR #278, opened 2021-01-11
Web config optimization

dukess:
Compatibility with older versions is preserved but users must update their config files. See below.

removed text field with "own BSB address".
fields "Own LPB address" and "Dest LPB address" renamed to "Own BSB/LPB address" and "Dest BSB/LPB address".

In config file:
options own_lpb_address and dest_lpb_address was renamed to own_address and dest_address.

Options own_bsb_address and dest_bsb_address was deleted.

@fredlcore To your judgment.

<s>Handler in Web config for dest_bsb_address added.</s>


1coderookie:
Question: With the current version where there are two independent sectors for BSB and LPB address - would it be possible to fill in different addresses?
E.g.: 
BSB: own 0x42, dest 0x00
LPB: own 0x32, dest 0x01
?
If so, then that could be a nice solution to preset different addresses which will be available one changes the bustype via Webconfig. I don't see that possibility if both systems would be merged within the two fields 'own address' & 'destination address' now.

dukess:
> With the current version where there are two independent sectors for BSB and LPB address - would it be possible to fill in different addresses?

Yes, own addresses can be set independently. 
But dest address allowed for LPB only.
If independent settings needed, then we will need to add another field to the EEPROM/Web config.

1coderookie:
But didn't you guys recently just added also the possibility for different destination addresses at BSB?
/<x>!<addr>
Cuz we also have different bus members at BSB and one probably would like to only query/set an EM for example.

dukess:
>But didn't you guys recently just added also the possibility for different destination addresses at BSB?

Yep.

For my clarification: how likely is it that the user will work in both (switching often) LSB and BSB modes at the same time?
If this scenario is possible then we need independent settings for both buses, sure.

1coderookie:
I can't say how likely it is, but I can imagine that some users will do that.
Could also be nice for us in testing scenarios, where we are plugging around between BSB&LPB.
I'd say: If it doesn't take too much memory or so, leave it as it is, at least right now we have the clear distinction between the settings of the different bustypes.

dukess:
roger!

1coderookie:
Great, thanks!

fredlcore:
Not so fast :)!
I couldn't think of any realistic situation where users would be switching between LPB and BSB, and even then own and destination address are likely to be the same (that's why we chose 0x42 and 0x00 respectively).
More variables lead to more potential confusion, both in the config as well as in the webinterface.
That's what we used to have before as well with just one bus(x,y,z) command in the config.
I would therefore go for one `own_address`and one `dest_address`.

dukess:
I turn my couch around! :-)

fredlcore:
Thanks! @1coderookie, can you adjust the manual accordingly? That would be great!

1coderookie:
Sure!

PR #280, opened 2021-01-12
Code for EEPROM erasing at startup (prototype).

dukess:
Code for EEPROM erasing at startup when pin 51 and GND shorted.
fix: disabling ETag support for /JC & /JK commands with parameters which will be send through POST
little bit of translation.


localization/LANG_EN.h
```diff
@@ -137,24 +137,41 @@
...
-#define STR516_TEXT "Standard values TSP heating circuit 1"
-#define STR536_TEXT "Standard values TSP heating circuit 2 "
-#define STR556_TEXT "Standard values TSP3"
-#define STR576_TEXT "Standard values TSP 4"
+#define STR500_TEXT "Mo"
+#define STR501_TEXT "Tu"
+#define STR502_TEXT "We"
+#define STR503_TEXT "Th"
+#define STR504_TEXT "Fr"
+#define STR505_TEXT "Su"
```
1coderookie, line 145: should be Saturday (Sa) here

localization/LANG_EN.h
```diff
@@ -137,24 +137,41 @@
...
+#define STR506_TEXT "Sa"
```
fredlcore, line 146: Shouldn't 505 be "Sa" and 506 "Su"?

1coderookie, line 146: should be Sunday (Su) here

localization/LANG_EN.h
```diff
@@ -137,24 +137,41 @@
...
+#define STR516_TEXT "Standard values TSP"
```
1coderookie, line 147: Suggestion: TSP may be better written out in full name "time switch program"?

localization/LANG_EN.h
```diff
@@ -137,24 +137,41 @@
...
+#define STR639_TEXT "Holiday period 4: last day"
+#define STR640_TEXT "Holiday period 5: first day"
+#define STR641_TEXT "Holiday period 5: last day"
+#define STR642_TEXT "Holiday period 6: first day"
+#define STR643_TEXT "Holiday period 6: last day"
+#define STR644_TEXT "Holiday period 7: first day"
+#define STR645_TEXT "Holiday period 7: last day"
+#define STR646_TEXT "Holiday period 8: first day"
+#define STR647_TEXT "Holiday period 8: last day"
 #define STR648_TEXT "Holiday operating level HC1"
```
1coderookie, line 164: HC1 needs to be removed here at STR648_TEXT also if the main names/description should be used for either HC1, HC2 and HC3

1coderookie (changes requested):
some small corrections for the LANG_EN-changes that have been made

localization/LANG_EN.h
dukess: It can be my mistake

localization/LANG_EN.h
dukess: Thanks! Corrected.

1coderookie (changes requested):
For 648 we could/should probably add 'heat circuit' or 'HC' so that it's clear to what 'holiday operating level' refers to (HC or DHW)..? (Sorry, didn't think of that before..)

dukess:
STR648_TEXT used in 3 categories (HC1 - HC3)

fredlcore:
I think we can leave out HC1-HC3 etc because the category already says what this refers to. We (try to) do this with all labels which are the same across different categories. This makes not only translation easier, but also saves memory.

fredlcore:
Or, if it's just a generic term like "heat circuit" then it should be fine. But we couldn't use it for DHW or other categories then.

1coderookie:
That's what I meant with 'heat circuit' / 'HC', because afaik there isn't a referring parameter/description for DHW for the holidays. But I'm fine with the decision to just leave it out completely here.

dukess:
@fredlcore @1coderookie  Should I fix something else here?

fredlcore:
Apart from the last decision, I think everything is clear, @1coderookie ?

1coderookie:
> Apart from the last decision, I think everything is clear, @1coderookie ?

Yes :)

BSB_lan_EEPROMconfig.h
```diff
@@ -107,19 +108,20 @@ typedef struct {
 PROGMEM_LATE const category_list_struct catalist[]={
   {CCAT_GENERAL,        CAT_GENERAL_TXT},
   {CCAT_IPV4,           CAT_IPV4_TXT},
-  {CCAT_MQTT,           CAT_MQTT_TXT}
+  {CCAT_MQTT,           CAT_MQTT_TXT},
+  {CCAT_BUS,            CAT_BUS_TXT}
 };
```
1coderookie, line 113: Oh great, thanks! I can think about the category names in German and English and make a list, so that you can add them? 
Would be much better than me messing around here.. ;)

1coderookie:
Cuz I dunno if my comment is obvious about the added Bus-category: Would it be ok for you guys if I think about it and make a list for the German and English categories which should/could be added to the table? I could post it in the other issue then and you guys could think about and maybe add them?

BSB_lan_EEPROMconfig.h
dukess: Sure!

fredlcore:
Yes, please do!

BSB_lan_EEPROMconfig.h
1coderookie: For 'Sensors' (=English) it should be 'Sensoren' in German btw, but I'll make the list for the categories and the German/English expressions, so that you can do it all in one process.

PR #281, opened 2021-01-12
Fix MDNS_HOSTNAME undefine

Ingramz:
It was added recently, but undefine referring to wrong define meant that MDNS still took space on Mega2560 by default.

With custom defs optimization and this change there was enough space to also fit MQTT in my case without changing anything else - good result.


fredlcore:
Thanks, yes, I changed the name but forgot the AVR-Section ;)...

PR #284, opened 2021-01-13
German translations for webconfig entries & categories 

1coderookie:
LANG_DE -> webconfig: 
- Added categories, so that we can use same names in second columns for LANG_EN also (like "Pins")
- Adjusted the names/expressions

LANG_EN -> webconfig:
- Added categories and names, but only where the German version didn't fit (please check if I did it right!)
- Adjusted the names/expressions

In general:
In both language files I use certain terms more than once for different parameters, so that we can put it once somewhere and then get them from there (e.g. Pins, Usage, Devices, Verwenden etc.) - I just don't know how to do that.. ;)

Note @dukess: I _think_ you already added the category 'Sensors' somewhere - that one can be erased now again cuz I decided to go with category names like OneWire and DHT, so that we can use the same names in the second colums like 'Pins', 'Devices' etc.  

Hope you guys are fine with it, changes will occur later probably again though ;)

For today: afk - have a nice one!


fredlcore:
Thanks, @1coderookie! However, I see a few problems here:
1. There are quite a few doublettes, for example "Verwenden". If this is the case, then we only make one definement. If it should still have a differently named definement, then the syntax would be `#define TXT_MARK_ABC TXT_MARK_XYZ`.
2. You have committed files with a tilde (~) at the end. These are seemingly some kind of temporary files. You need to commit the "real" files (i.e. LANG_DE.h for example).
3. I don't really understand where you put the categories, because now you have things like "OneWire: Pins", so "OneWire" is part of the description, not the category?

fredlcore:
Ok, the tilde filename is now fixed, but doublettes are still around ;)...

1coderookie:
> Thanks, @1coderookie! However, I see a few problems here:
> 
> 1. There are quite a few doublettes, for example "Verwenden". If this is the case, then we only make one definement. If it should still have a differently named definement, then the syntax would be `#define TXT_MARK_ABC TXT_MARK_XYZ`.

That was 'on purpose' and was what I mentioned in my edited comment above: I put the same names/descriptions to the definements, but I (still ;) ) don't really know where and how to put them so that each entry just grabs it from that one source. Could you please take care of that? I'll have a look at it afterwards, maybe then I'll understand.. 

> 2. You have committed files with a tilde (~) at the end. These are seemingly some kind of temporary files. You need to commit the "real" files (i.e. LANG_DE.h for example).

Yes, that was quite irritating and confusing for me also, but I thought that it maybe had to be (diff-files?) that way, cuz they just appeared within my fork suddenly and were uploaded here in the PR also. I just deleted them manually now from my fork online, so I hope it's ok now?!?

> 3. I don't really understand where you put the categories, because now you have things like "OneWire: Pins", so "OneWire" is part of the description, not the category?

Maybe that was in one of the 'wrong' ~files? I just had a look at the LANG-files again and I think it's not in there anymore. The added categories are underneath the definements for the descriptions, right where the other categories are defined/listed also.
In other words: if I click on the [last comment](https://github.com/fredlcore/bsb_lan/pull/284/commits/f84110a0c884c6b06032ff0d9a9817093589b24e)  and igrnore looking at the ~-files, then the right LANG-files are displayed.

1coderookie:
Sorry for the confusion and the hustle with my PR and the doublettes, it's really new for me :(

fredlcore:
I think it would really help if you would remove the doublettes yourself - it's really easy and you know better where they might be (also for future changes). Here's an example:
You wrote:
```
#define CF_AVERAGESLIST_TEXT "Parameter"
#define CF_CURRVALUESLIST_TEXT "Parameter"
```
Both are the same and you want to use both definements (which is perfectly fine because it is much easier to understand in the code dealing with the "current values list" when the variable is named "CF_CURRVALUESLIST_TEXT" and not "CF_AVERAGESLIST_TEXT" (which is from a completely different context).
So, in order to save memory and also work for other translators, you just write:
```
#define CF_AVERAGESLIST_TEXT "Parameter"
#define CF_CURRVALUESLIST_TEXT CF_AVERAGESLIST_TEXT
```
That tells the compiler that whenever it comes across `CF_CURRVALUESLIST_TEXT`, it will take the contents of `CF_AVERAGESLIST_TEXT` which in turn will be "Parameter".
That's all :).
If we later decided at some point to rename that term, we can just do it in one place - or, if we realize that there should be two different words at a later stage, we can just tell "CF_CURRVALUESLIST_TEXT" to be something else.

fredlcore:
And don't worry, you're doing great with the PRs! It will be a real time-saver once you have done it a few times, I'm speaking from my (not too old) experience in that matter :)...

1coderookie:
Thanks for building me up.. ;)

Ok, I'll try tomorrow with the doublettes, it does sound quite logic and not too complicated to implement! _..fingers crossed.._ ;) 

Need to get off the computer again for now, the heating system is already sleeping and the room is too cold - and actually I'm too tired in my brain also.. :P

fredlcore:
Good idea (for me too ;) )!

1coderookie:
Hope it's fine now :)

fredlcore:
Looks great, thanks!

PR #285, opened 2021-01-14
New categories for Web config added

dukess:
New categories for Web config added according to @1coderookie
text files will be cached for 60 seconds.
fix: configure LED output before first usage
Last (?) RAM consumption optimization.
Optimized text resources in web config
Bug in ETag checking fixed.


fredlcore:
I merged @1coderookie's changes first, so I can't merge now, could you check for conflicts? Thanks!

fredlcore:
CI says:
```
In file included from /tmp/tmpe6juy50x/src/BSB_lan.ino:441:0:
src/BSB_lan_defs.h:760:43: error: 'CF_LOGAVERAGES_TEXT' was not declared in this scope
 const char CF_LOGAVERAGES_TXT[] PROGMEM = CF_LOGAVERAGES_TEXT;
                                           ^~~~~~~~~~~~~~~~~~~
src/BSB_lan_defs.h:760:43: note: suggested alternative: 'CF_LOGAVERAGES_TXT'
 const char CF_LOGAVERAGES_TXT[] PROGMEM = CF_LOGAVERAGES_TEXT;
                                           ^~~~~~~~~~~~~~~~~~~
                                           CF_LOGAVERAGES_TXT
```

dukess:
Fixed. I made some changes in LANG files.

fredlcore:
Thanks! I have changed the sort order a bit so that entries of the same category are next to each other.
One last thing for the config page: I think it would be better to move the entries to the top and the "debug infos" below that. Because most users won't even know what protected pins are, and some entries are redundant (such as pins), so it's not really necessary (at the top). We could even think of not displaying those entries that also exist in the web config (if web config is enabled).

1coderookie:
Great, thanks for the sorting, that was the next thing I wanted to ask about (how to do it). ;)

Yes, moving the webconfig-table itself up to the top would be good. But please don't get rid of it in case the webconfig is enabled, cuz there are informations within that listing which aren't available in the webconfig.

@dukess: 
Just did a quick check of the current version: OneWire-Geräte/Devices is missing in the webconfig now (I _think_ it was also present before if no sensors are connected)..

fredlcore:
I have OneWire pins as a setting option here in Webconfig...

dukess:
@1coderookie I deleted it because it never used. Long time ago i planned add OneWire devices list to config but drop this idea as useless.

1coderookie:
@fredlcore: Yes, PINS - but there was an entry where the devices were listed.

@dukess: Oh, I thought that was an automatic output of the connected devices.. So you planned it like for the MAX!-devices? You're right, I guess that would be useless cuz one doesn't has to add them manually within the config due to the autodetection..

dukess:
Yes, something like that

PR #294, opened 2021-01-18
LANG_DE parameter names: removed abbreviations, fixed typo, renaming …

1coderookie:


fredlcore:
Thanks!

localization/LANG_DE.h
```diff
@@ -488,23 +488,23 @@
 #define STR2452_2_TEXT "Reglerverzögerung Gebläseleistung"
 #define STR2452_3_TEXT "Reglerverzögerung Gebläsedrehzahl"
 #define STR2453_TEXT "Reglerverzögerung Dauer"
-#define STR2454_TEXT "Schaltdifferenz Ein HK's"
-#define STR2455_TEXT "Schaltdifferenz Aus Min HK's"
-#define STR2456_TEXT "Schaltdifferenz Aus Max HK's"
-#define STR2457_TEXT "Einschwingzeit HK's"
+#define STR2454_TEXT "Schaltdifferenz ein HK"
+#define STR2455_TEXT "Schaltdifferenz aus Min HK"
```
fredlcore, line 492: I know I said it differently before, but in this case it might make sense to put the "Min." (with a dot for all abbreviations unless they are acronyms(?) like "HK" or "TWW") in front, because if there is a word after the "Min." it's not quite clear whether this applies to "Schaltdifferenz" or "HK" (as if one could set a range of HKs with a minimum and a maximum number)

1coderookie: Hmm, I didn't change the order of the words in this case, but I get what you mean though. 
Just looked up one manual really quick (programming handbook) and there it's even named different(?!):
2454 Schaltdifferenz Kessel
2455 Schaltdiff Kessel Aus Min
2456 Schaltdiff Kessel Aus Max
So besides the fact that Min&Max are also w/o a "." at the end and that it's standing completely at the end, I just wonder where the different names from the LANG_DE entries come from, cuz up to me there is/could be a difference in the meaning of 'Schaltdifferenz Kessel' and 'Schaltdifferenz HK'..
I'll write myself a note about that and will have a look at different manuals as soon as I have a little more time for that, ok? (Like to finish the chap5.1 etc before..)
But anyway (either it's Kessel or HK), I'd agree that it would be better to understand if it would be written different - but then I'd vote for putting the Min/Max at the end and not at the beginning, cuz a) I _think_ with the other entries it's made the same like that and b) if you look at the three parameters within one query you see Schaltdifferenz always at the same first position. And it's also really clear to me, that it's the Schaltdifferenz-Maximum.
So _if_ somehow HK is correct, then I'd vote for "Schaltdifferenz HK ein Min". But of course I can also change it and put the Min/Max at the beginning if you really prefer that one - I'm fine with either variant - it's just a lot more work to do if I have to set Min/Max to the beginning for all entries though..

fredlcore: Yes, whether it's at the end or beginning doesn't matter to me, both ways are clearer than the present one. And it's not something that you have to do (or do it now), it's just something that I would consider. As for "Min/Max" without a dot, it's just that it's the grammar nazi in me speaking, because a dot makes it clear that it's an abbreviation and not a different term. But nothing that's urgent...

1coderookie (commented):
Upps, did I press awrong button? Didn't mean to start a review, just wanted to add a comment!?!

PR #295, opened 2021-01-19
RGT emulator, "Presence" and "TWW push" buttons handler.

dukess:
New feature: RGT 1-3 emulator: user can set up to 5 sensors per HC with average calculation.
New feature: Add handling presence buttons and TWW push
Bugfix: receiving incomplete data after config submitting.


dukess:
New faeture was tested on "dry" Mega (without connection to heater)

BSB_lan.ino
```diff
@@ -597,6 +597,19 @@ unsigned long custom_timer = millis();
...
+#endif
+
+#ifdef BUTTONS
+volatile byte PressedButtons = 0;
+#define TWW_PUSH_BUTTON_PRESSED 1
+#define ROOM1_PRESENCE_BUTTON_PRESSED 2
+#define ROOM2_PRESENCE_BUTTON_PRESSED 4
+#define ROOM3_PRESENCE_BUTTON_PRESSED 8
+#endif
+
```
fredlcore, line 612: I would wrap these lines with #ifndef so that users could define them in _config.h, so that these settings could be defined there and stay the same even if a new version of BSB-LAN is flashed.

fredlcore: Or will these be overriden with the EEPROM configuration I just saw below?

dukess: Thse definitions for internal usage only (just for comfort reading).

fredlcore: Ok, great!

1coderookie:
@dukess: Just thought about something about the integration of the buttons: I was playing around with the integration of buttons in the past a bit and stumbled across the need to implement a debounce function within the code (something like check state -> delay -> check button state again -> execute command). 
Did you think of this and integrated it somehow? Otherwise problems could occur with bouncing of the button and therefore sending the command a few times even if one just pushed the button only once.

dukess:
Signal will send when button released. Multiple bounces can set flag 
multiple time but flag can be read and cleared in handler only one time per 
loop. Flag will cleared after 0,25 - 0,5 sec (1-2 telegrams) after entering 
to handler. I think that the rattling will stop during this time.
In short: we need testing. But if here appear a problem we can add 
additional delay to handler.
"U. D." <notifications@github.com> 20 января 2021 г. 13:38:53 написал:

1coderookie:
Ok, thanks for explanation. Just to be definitely on the safe side, I'd suggest to add that delay anyway - but I'll test it the way it is now of course (_maybe_ even today ;) ), but maybe it wouldn't be representative even if it would work out fine.. Just a thought.. ;)

fredlcore:
However, in the end, all that would happen is to send the same message twice or so over the bus.

1coderookie:
Ok, so no problem at all?! Great :)

1coderookie:
So, I did a test with TWW and RGT1 buttons.
For TWW push it seems to work:
```
14:34:18.476 -> setting line: 1603 val: 01 01 
14:34:18.675 -> LAN->HEIZ SET 1603 Trinkwasser - Manueller TWW-Push: ---
14:34:18.675 -> DC C2 00 0D 03 3D 31 05 73 01 01 74 B4 
14:34:18.675 -> HEIZ->LAN ACK 1603 Trinkwasser - Manueller TWW-Push: 
14:34:18.675 -> DC 80 42 0B 04 31 3D 05 73 60 07 
14:34:19.106 -> HEIZ->ALL  INF 1602 Trinkwasser - TWW Status: 01000111
14:34:19.139 -> DC 80 7F 0E 02 31 00 02 12 01 00 47 F6 35 
14:34:19.139 -> INF: TWW-Status: 71

```
But for RGT1 button it seems that there's only the query but no set-command afterwards:
```
14:32:28.819 -> LAN->HEIZ QUR 10102 Benutzerdefiniert - INFO HK1 - TBD: 
14:32:28.819 -> DC C2 00 0B 06 00 2D 02 11 2C B0 
14:32:28.819 -> HEIZ->LAN ANS 10102 Benutzerdefiniert - INFO HK1 - TBD: 01 01 1E 84 FF FF FF FF 00 05 
14:32:28.819 -> DC 80 42 15 07 2D 00 02 11 01 01 1E 84 FF FF FF FF 00 05 9A 02 
14:32:28.819 -> #10102:  01011E84FFFFFFFF0005 - unknown type
```
The mode doesn't change.

dukess:
Good news :)
Problem is clear: i forgot that 10102-10104 is VT_UNKNOWN and 
decodedTelegram.value is empty. Here is two way: define programs 
10102-10104 as VT_STRING (or add new type like VT_DUMP) or use 
decodedTelegram.telegramDump for parsing.


@fredlcore what you prefer?

"U. D." <notifications@github.com> 20 января 2021 г. 16:41:40 написал:

dukess:
@1coderookie 
Fast fix  with decodedTelegram.telegramDump using
 https://github.com/fredlcore/bsb_lan/pull/296

1coderookie:
Will test when PR is merged, thanks

dukess:
Sorry. Was wrong and fast.

fredlcore:
Since VT_UNKNOWN is a valid data type which contains the dump of the payload, I would treat it as either as string or as byte array - don't know if this is easily possible? In any case, I will create a new parameter soon for that specific one, but if telegramDump could easily be added and contain the payload of any decodedTelegram, then that would certainly be a useful addition.

dukess:
Fixed: sscanf has wrong format. Tested with standalone sketch.

PR #305, opened 2021-01-22
Added comment about MAC address, added and adjusted entries in webconfig 

1coderookie:
- Added comment in config.h about MAC address -> LAN only, doesn't apply to ESP
- Added and adjusted entries for/in webconfig: "MAC address (LAN only)" and "WLAN password"


1coderookie:
@dukess: Can you please review this? I tried locally and it works fine, just want to make sure that I really did it correct.. Thanks!

dukess:
Looks fine!

PR #317, opened 2021-01-30
Fix trusted_ip check

torbennehmer:
The check against the trusted IP Addresses did a direct memory
comparison, which is dependent on compiler and memory layout.
Changed the implementation to use the overloaded == operator from
IPAddress.h, which works regardless of this. Note, that != is not
overloaded, hence the explicit negation, != wouldn't work.


fredlcore:
Thanks!

PR #319, opened 2021-01-31
Fix for DHT22

dukess:
Saving state between requests to avoid early repeated reading of the sensor.


1coderookie:
Works! Thanks!

PR #323, opened 2021-02-02
Fix BlueDot BME280 library for ESP32 MCU + LANG_RU update

dukess:
Fix BlueDot BME280 library for ESP32 MCU


fredlcore:
Thanks for these corrections! Could you submit them to the BME280 repository as well? Then we would know we could safely update to any future version.

dukess:
First we need to ask @1coderookie to test fix with real sensor.  If fix will work then i can make PR.

1coderookie:
Will do, probably this afternoon. 
Just do get it right: Can I test it with my previous setup or do I have to test it with the ESP32 setup? ESP32 I guess by the comment - right?

dukess:
With your previous setup.

dukess:
But if you can do test with ESP32 - it would be great

1coderookie:
If it compiles w/o problems using PlatformIO, I'll try. One sensor is enough? Would like to avoid cutting&soldering for changing the address of the second one just for a test. But if it would be neccessary to test it with two sensors, I'll do it of course..

dukess:
>If it compiles w/o problems using PlatformIO, I'll try.

If not - it can be leaved for next time.

>One sensor is enough?

One will be enough. But if you have a time, opportunity, and the need for a second sensor, then such testing will certainly be great.

Thank you!

1coderookie:
Ok, so I'll test it with one sensor first. Hope that one is still ok - I just twisted the 3,3V and GND wires and connected them wrong by accident, few seconds later that thing got smelly and the sensor itself got SO hot that I just burned myself when I touched it..grrrrrrr.. :(

dukess:
:(

1coderookie:
WOW! I didn't expect that they are THAT solid - it's still working! 
Note: If one wants to dry up the sensor, just wire it wrong for a few seconds - when it gets hot, it's not a bug, it's a feature.. ;)

Works at ESP32:
Bildschirmfoto von 2021-02-02 13-49-40

Works at Due using WiFi-ESP8266-solution:
Bildschirmfoto von 2021-02-02 14-00-43

dukess:
> If one wants to dry up the sensor

You're lucky today. You should buy lottery ticket :)

fredlcore:
This is OT, but since this matter is closed just a heads up for you guys: I have restructured the directory layout and moved (almost) everything below the new folder BSB_LAN. This makes it easier for new Arduino IDE installations because they do not have to rename the folder and can just double-click on the "BSB_lan.ino" (after they have renamed BSB_lan_config.h).
Also, the main page now looks much less cluttered and you get to read the README.md without having to scroll down endlessly ;).

The reason I did this was that I had hoped I could move the libraries that conflict with the ESP32 to an extra folder, but this does not work because the Arduino IDE only compiles libraries under "src".

dukess:
You need to ajust paths and workflows.

dukess:
I think we can place "schematics" and "localization/www" and "localization/json" on top level:
/BSB-LAN
/localization/json
/localization/www
/schematics
...

1coderookie:
Good idea, but then the main folder must be named "BSB_lan" as well (same name as the .ino-file) and not "BSB_LAN".

fredlcore:
For schematics, I'd agree, localization is code-related and thus could stay below BSB_LAN.
@1coderookie: At least here on my Mac it works (seems to be case-insensitive).

1coderookie:
Here it doesn't work, IDE is complaining about the name..

fredlcore:
Strange, Mac has a case-sensitive filesystem, too. Ok, then I'll take the opportunity and rename the files to BSB_LAN_... because it never really made sense to me why BSB was written in caps and lan in small caps...

1coderookie:
Good point, didn't get that either.. ;)
So I'll adjust the paths for the images in readme-files right now, using "BSB_LAN".

fredlcore:
Yes, thanks!

PR #325, opened 2021-02-02
FIX ESP32 EEPROM Emulation

do13:
This commit fixes ESP32 EEPROM class. Issue #322


fredlcore:
Unfortunately, it won't solve the problem.

PR #327, opened 2021-02-03
improve options for MQTT

Luposoft63:
-new option for query parameters from heater via MQTT
-all the code wich is for sending mqtt-messages came in a new function mqtt_sendtoBroker
 (this code was not changed except log_parameters[i] to param)
-sending ACK to broker moved prior set-command (it looks better in the log, when first ACK and then the result)
-after set-command run mqtt_sendtoBroker to send new value immediately

Now I hope, this PR runs better as my first...

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -9094,21 +9094,19 @@ return false;
 #ifdef MQTT
 void mqtt_callback(char* topic, byte* payload, unsigned int length)
 {
-  boolean setcmd;
+  //boolean setcmd;
   printlnToDebug(PSTR("##MQTT#############################"));
   printToDebug(PSTR("mqtt-message arrived ["));
   printToDebug(topic);
   printlnToDebug(PSTR("] "));
   char C_value[24];
   strcpy_P(C_value, PSTR("ACK_"));   //dukess
+  char firstsign;
+  firstsign=' ';
   switch ((char)payload[0])
   {
-    case 'I':setcmd=false;break;
-    case 'S':setcmd=true;break;
-    default:
-      printlnToDebug(PSTR("mqtt_callback: missing 'I' or 'S' at start"));
-      return;
-      break;
+    case 'I':{firstsign='I';printToDebug(PSTR("I"));break;}
+    case 'S':{firstsign='S';printToDebug(PSTR("S"));break;}
   }
   //buffer overflow protection    //dukess
   if(length > sizeof(C_value) - 4)
@@ -9122,12 +9120,8 @@ void mqtt_callback(char* topic, byte* payload, unsigned int length)
   }
   C_value[length+4]='\0';
   char*C_payload=C_value+ 4;  //dukess
-  C_payload++;
+  if(firstsign!=' ') C_payload++; //skip I/S
   int I_line=atoi(C_payload);
-  C_payload=strchr(C_payload,'=');
-  C_payload++;
-  printFmtToDebug(PSTR("%с%d=%s \r\n"), setcmd?'S':'I', I_line, C_payload);
-  set(I_line,C_payload,setcmd);  //command to heater
   String mqtt_Topic;
   if(MQTTTopicPrefix[0])
   {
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks, before I look into this in more detail, please check indention and adhere to the bracket-style of the rest of the code (i.e. curly brackets in the same line as the if/for/while command). Also, please say a bit more as to why the changes are necessary and/or an improvement and what they do. Otherwise we don't know how to include these in the manual and no one will be using them...

Luposoft63:
Thanks for your comment,
actually we can send a command from broker to bsb-lan to switch a paramter at the heater via MQTT.
bsb-lan returned an ACK only.
User asked for a possibility to send a query via MQTT as well.
with this PR:
-we can send from broker e.g. 'set mqtt2Server publish BSB-LAN 700'
the heater returns an ACK and the value.
-when we send a switch command (e.g. Sxxxx=xxx), bsb-lan will send an ACK to broker, set the parameter at the heater and then return the new value to the broker
 
We have a full integration of MQTT.
Using MQTT with its small telegram offers a great advantage.
In my opinion, this could peel of the need of using HTTPMOD.

fredlcore:
Thanks, that really sounds good! @dukess: Could you also have a look at it again? Your comments last time were really helpful.
There are still a few indention errors and inconsisten use of else without brackets or brackets sometimes in extra line and sometimes in same line.

BSB_LAN/BSB_LAN.ino
```diff
@@ -8983,6 +8906,105 @@ uint8_t pps_offset = 0;
   update_server.handleClient();
 #endif
 } // --- loop () ---
+
```
dukess, line 8909: It's a good idea to move MQTT in a separate function!
Main loop will looks more comfort for reading

BSB_LAN/BSB_LAN.ino
```diff
@@ -9069,51 +9091,52 @@ return false;
...
-    case 'S':setcmd=true;break;
-    default:
-      printlnToDebug(PSTR("mqtt_callback: missing 'I' or 'S' at start"));
-      return;
-      break;
+  char firstsign;
+  firstsign=' ';
+  switch ((char)payload[0]) {
+    case 'I':{firstsign='I';printToDebug(PSTR("I"));break;}
+    case 'S':{firstsign='S';printToDebug(PSTR("S"));break;}
```
dukess, line 9106: case 'I':
    case 'S':firstsign=(char)payload[0];printFmtToDebug(PSTR("%с"), firstsign);break;

it can be shorter... or not. %)

BSB_LAN/BSB_LAN.ino
```diff
@@ -9069,51 +9091,52 @@ return false;
...
+    mqtt_sendtoBroker(I_line);  //send mqtt-message
+  }
+  else { //command to heater
+    C_payload=strchr(C_payload,'=');
+    C_payload++;
+    printFmtToDebug(PSTR("%d=%s \r\n"), I_line, C_payload);
+    set(I_line,C_payload,firstsign=='S');  //command to heater
+    mqtt_sendtoBroker(I_line);  //send mqtt-message after command
+  }
   printlnToDebug(PSTR("##MQTT#############################"));
```
dukess, line 9140: Little bit shorter.
```
  if (firstsign==' ') { //query
    printFmtToDebug(PSTR("%d \r\n"), I_line);
  } else { //command to heater
    C_payload=strchr(C_payload,'=');
    C_payload++;
    printFmtToDebug(PSTR("%d=%s \r\n"), I_line, C_payload);
    set(I_line,C_payload,firstsign=='S');  //command to heater
  }
  mqtt_sendtoBroker(I_line);  //send mqtt-message after command
```

Luposoft63: done

BSB_LAN/BSB_LAN.ino
Luposoft63: printFmtToDebug(PSTR("%с"), firstsign)
this won't work, the output is 'c'
I also tried printFmtToDebug("%с", firstsign). But the same issue.
why that?

dukess: Sorry. 'c' in PSTR("%с") not latin 'c', it is russian 's'. it have different ASCII/UNICODE code but looks similar.
Correct version: 
`PSTR("%c")`

Luposoft63: confusing:
I say: printFmtToDebug(PSTR("%с"), firstsign) don't work

you say: PSTR("%c") is correct

where is the difference?

dukess: Just delete 'c' in PSTR("%с") and type 'c'.
 'c' in latin alphabet and 'с' in cyrillic alphabet  is very similar but it is different letters with different ASCII/Unicode codes.

fredlcore:
Thanks @dukess for the input and @Luposoft63 for the changes made! @Luposoft63, please let @1coderookie know about any changes that should be added to the manual so that people can make good use of it.

PR #332, opened 2021-02-06
Mqtt last will topic handling & code improvements

torbennehmer:
This PR adds a [last will/status](https://www.hivemq.com/blog/mqtt-essentials-part-9-last-will-and-testament/) topic handling. This way, we always have the current state of the MQTT device available at the broker. Tools like Home Assistant (HA) use this to determine, if the integration in question is available or currently down. The will/status subtopic used is "status" below the configured prefix, f.x. "BSB-LAN/status". Values can be "online" and "offline". 

To make this easier, I took the liberty to refactor the code a bit:

Improved the mqtt_connect code. We allocate a bit less RAM on each call (as we're connected almost all of the time). When connecting, we register our last will & testament at the status topic. After a successful connection, we update the status topic to online and we're good.

Extracted the (duplicated) disconnect code and centralized it into mqtt_disconnect. When disconnecting, we set our status topic to offline and cleanup afterwards. Added logging to this location as well. We call the disconnect upon board reset as well so that we exit the MQTT connection gracefully and go offline immediately and not via timeout.

Created mqtt_get_will_topic to centralize will topic parsing. This should make it easy to make it configurable later on.

An excess mqtt_connect call has been removed from mqtt_sendtoBroker. Once we enter the latter call we can safely assume that we're already connected. loop checks this already beforehand, and if we come out of the callback, we are connected by definition as well.

Stramlined disconnect handling in the MQTT calls in the loop to break it down to a single mqtt_disconnect.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -9172,7 +9172,7 @@ return false;
  *   none
  */
 #ifdef MQTT
-const String& mqtt_get_will_topic() {
+const String mqtt_get_will_topic() {
   // Build (Last) Will Topic
   String MQTTLWTopic = "";
   if (MQTTTopicPrefix[0]) {
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -4220,12 +4220,7 @@ bool SaveConfigFromRAMtoEEPROM() {
 #ifdef MQTT
         case CF_MQTT:
         case CF_MQTT_IPADDRESS:
-          if (MQTTPubSubClient) {
-            delete MQTTPubSubClient;
-            MQTTPubSubClient = NULL;
-            mqtt_client->stop();
-            delete mqtt_client;
-          }
+          mqtt_disconnect();
           break;
 #endif
         default: break;
@@ -6244,7 +6239,8 @@ void transmitFile(File dataFile) {
 
 /** *****************************************************************
  *  Function: resetBoard
- *  Does: restart Arduino
+ *  Does: restart Arduino, will disconnect from MQTT if connected,
+ *        required to send update the WILL topic correctly.
  *  Pass parameters:
  *   none
  * Parameters passed back:
@@ -6252,9 +6248,12 @@ void transmitFile(File dataFile) {
  * Function value returned:
  *   none
  * Global resources used:
- *   none
+ *   MQTT instance
  * *************************************************************** */
 void resetBoard() {
+#ifdef MQTT
+  mqtt_disconnect();
+#endif
   forcedflushToWebClient();
   delay(10);
   client.stop();
@@ -8661,19 +8660,13 @@ uint8_t pps_offset = 0;
       }
       if (MQTTPubSubClient != NULL && !mqtt_mode)  //Luposoft: user may disable MQTT through web interface
       {
-        if (MQTTPubSubClient->connected())
-        {
-          MQTTPubSubClient->disconnect();
-          printlnToDebug(PSTR("MQTT was disconnected on order through web interface"));
-        }
+        // Actual disconnect will be handled a few lines below through mqtt_disconnect().
+        printlnToDebug(PSTR("MQTT will be disconnected on order through web interface"));
       }
     }
   }
-  if (MQTTPubSubClient && mqtt_mode == 0) {
-    delete MQTTPubSubClient;
-    MQTTPubSubClient = NULL;
-    mqtt_client->stop();
-    delete mqtt_client;
+  if (mqtt_mode == 0) {
+    mqtt_disconnect();
   }
 #endif
 
@@ -9018,7 +9011,7 @@ uint8_t pps_offset = 0;
 void mqtt_sendtoBroker(int param) {
   // Declare local variables and start building json if enabled
   String MQTTPayload = "";
-  String MQTTTopic = "";
+  String MQTTTopic = ""; 
   if (mqtt_mode == 2 || mqtt_mode == 3) {
     MQTTPayload = "";
     // Build the json heading
@@ -9033,7 +9026,7 @@ void mqtt_sendtoBroker(int param) {
       MQTTPayload.concat(F("\":{\"id\":"));
   }
   boolean is_first = true;
-  if (mqtt_connect()) {              //Luposoft, new funct
+  if (mqtt_connect()) { // This call seems unneccessary
     if (is_first) {is_first = false;} else {MQTTPayload.concat(F(","));}
     if (MQTTTopicPrefix[0]) {
       MQTTTopic = MQTTTopicPrefix;
@@ -9041,7 +9034,7 @@ void mqtt_sendtoBroker(int param) {
     }
     else
       MQTTTopic = "BSB-LAN/";
-// use the sub-topic "json" if json output is enabled
+    // use the sub-topic "json" if json output is enabled
     if (mqtt_mode == 2 || mqtt_mode == 3)
       MQTTTopic.concat(F("json"));
     else
@@ -9116,13 +9109,7 @@ void mqtt_sendtoBroker(int param) {
 #ifdef MQTT
 boolean mqtt_connect()
 {
-  char* MQTTUser = NULL;
-  if (MQTTUsername[0])
-    MQTTUser = MQTTUsername;
-  const char* MQTTPass = NULL;
-  if (MQTTPassword[0])
-    MQTTPass = MQTTPassword;
-  if (MQTTPubSubClient == NULL)
+  if(MQTTPubSubClient == NULL)
   {
     mqtt_client= new ComClient();
     uint16_t bufsize;
@@ -9137,12 +9124,19 @@ boolean mqtt_connect()
   }
   if (!MQTTPubSubClient->connected())
   {
+    char* MQTTUser = NULL;
+    if(MQTTUsername[0])
+      MQTTUser = MQTTUsername;
+    const char* MQTTPass = NULL;
+    if(MQTTPassword[0])
+      MQTTPass = MQTTPassword;
     IPAddress MQTTBroker(mqtt_broker_ip_addr[0], mqtt_broker_ip_addr[1], mqtt_broker_ip_addr[2], mqtt_broker_ip_addr[3]);
     MQTTPubSubClient->setServer(MQTTBroker, 1883);
+    String MQTTWillTopic = mqtt_get_will_topic();
     int retries = 0;
     while (!MQTTPubSubClient->connected() && retries < 3)
     {
-      MQTTPubSubClient->connect(PSTR("BSB-LAN"), MQTTUser, MQTTPass);
+      MQTTPubSubClient->connect(PSTR("BSB-LAN"), MQTTUser, MQTTPass, MQTTWillTopic.c_str(), 0, true, PSTR("offline"));
       retries++;
       if (!MQTTPubSubClient->connected())
       {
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks! While I'm not using MQTT and therefore cannot go into the details of the implementation, it looks good to me. @Luposoft63 and @dukess, do you want to have a look at it since both of you have been working on MQTT so far?
For me, it's just a few cosmetic things, indention is not correct at least in the `query(param)` line, and the position of curly brackets is changing. I'll have to set up code-style-guidelines at some point I guess ;). For now, most of the code uses
```
if (...) {
...
} else {
...
}
```

torbennehmer:
@fredlcore I've fixed the indentations. The code formatting is still inconsistent, but I'd like to keep it that way for this PR or it gets cluttered to much. I'd rather take a few minutes to reformat the code in its own PR, this would make change tracking easier I guess.

Please cross check the indentation change I have in mqtt_sendtoBroker due to the removal of the excess call. I triple-checked it and got a missing upstream change that got lost during the rebase. It should be complete now, but I'd feel safer if somebody else validates this explicitly.

fredlcore:
Thanks!
As for the removal of the call, since it was not written by me, could you guys have a look, @dukess and @Luposoft63?

Luposoft63:
I'm prepared to test it tomorrow.
@torbennehmer 
I think about the test scenario:
while normaly work the status is allways online.
If I switch off the arduino it goes to offline.
If I switch on/off via webif it goes on-/offline.

Am I right? Are there another testing aspects?

dukess:
@fredlcore @torbennehmer 
Hi! I have about the same problem: I don't use MUTT, so I won't be able to check the correctness of the code and its compatibility in real situation but according to the description, the improvements look great.

Regarding the design of the code: can we use a tool like Uncrustify for automatic formatting?

torbennehmer:
@Luposoft63 That's the general idea. The status topic should give you the current state of the BSB-LAN Adapter. I tested it by using mosquitto_sub to watch BSB-LAN/status. Once the chip has booted you'll get an online message. If you reset the board by any way that hits mqtt_disconnect (including the /N URL method), you should instantly get an offline notification. In case you're flasshing or simply shutting it off ungracefully, you should also get an offline message by way of the last will message registered upon connection. Be aware, the the last case involves a timeout with your MQTT broker, my mosquitto installation needs roughly 2 minutes until it realizes, that the BSB-LAN went down ungracefully.

fredlcore:
@Luposoft63: Great, thanks for testing this!
@dukess: Uncrustify looks great, should this be run via CI or on each contributor's computer?

dukess:
@fredlcore you can run it time to time on your PC. (I tried to config Uncrustify at weekend but it contain to many parameters. The result was not bad, but I did not like the formatting of some elements, and I did not find how to configure them.)

BSB_LAN/BSB_LAN.ino
```diff
@@ -9143,7 +9143,8 @@ boolean mqtt_connect()
       }
       else
       {
-        printlnToDebug(PSTR("Connect to MQTT broker"));
+        printlnToDebug(PSTR("Connect to MQTT broker, updating will topic"));
+        printFmtToDebug(PSTR("Will topic: %s"), MQTTWillTopic.c_str());
```
Luposoft63, line 9144: please add \r \n for new line
printFmtToDebug(PSTR("Will topic: %s\r\n"), MQTTWillTopic.c_str());

BSB_LAN/BSB_LAN.ino
```diff
@@ -9200,10 +9201,14 @@ String mqtt_get_will_topic() {
 #ifdef MQTT
 void mqtt_disconnect() {
   if (MQTTPubSubClient) {
-    printlnToDebug(PSTR("Disconnect from MQTT broker, updating will topic"));
     if (MQTTPubSubClient->connected()) {
-      MQTTPubSubClient->publish(mqtt_get_will_topic().c_str(), PSTR("offline"), true);
+      String MQTTWillTopic = mqtt_get_will_topic();
+      printlnToDebug(PSTR("Disconnect from MQTT broker, updating will topic"));
+      printFmtToDebug(PSTR("Will topic: %s"), MQTTWillTopic.c_str());
```
Luposoft63, line 9202: @torbennehmer 
printFmtToDebug(PSTR("Will topic: %s\r\n"), MQTTWillTopic.c_str());  as well

Luposoft63:
I'm  using the inbuild mqtt from fhem.
So we have it tested at two different broker.
It works fine and I love it.
Thanks for contribute

1coderookie:
@torbennehmer : Please let me know if (and how) any of these functions should be mentioned in the MQTT chapter of the manual as well or if it's something that works 'under the hood' and that the end-user probably doesn't need to know about/be aware of. (In case I should mention it: would be better maybe to talk about that via Email..)
Thanks

fredlcore:
@Luposoft63: Thanks for testing! I just saw that @hacki11 is also using MQTT for his plugin, so maybe this could be of interest for you, too?
@torbennehmer: Once the changes @Luposoft63 requested are included, I'll merge your PR. Thanks again!

BSB_LAN/BSB_LAN.ino
torbennehmer: A damn, I saw that in my logs but forgot to fix it. you'll get an updated PR in a minute.

BSB_LAN/BSB_LAN.ino
torbennehmer: Same here

torbennehmer:
> I'm using the inbuild mqtt from fhem.
> So we have it tested at two different broker.
> It works fine and I love it.
> Thanks for contribute

You're welcome! It was a log of fun writing it (normally I work with GBs of RAM ;-))


> @torbennehmer : Please let me know if (and how) any of these functions should be mentioned in the MQTT chapter of the manual as well or if it's something that works 'under the hood' and that the end-user probably doesn't need to know about/be aware of. (In case I should mention it: would be better maybe to talk about that via Email..)

I would mention it, as there currently is no magic behind it. Just drop me a quick hint via torben <at> nehmer <dot> net (in German or English) and I'll get in touch with a few explanaitions and an example config for Home Assistant.

Best regards, 
Torben

torbennehmer:
@hacki11 Right now, the system will only watch the configured IDs, which works well for me. Moving this to a more dynamic subscription mechanism would be possible I think.

hacki11:
Ok that's a show stopper for me. The user should not need to change headers, recompile, reflash just to get a new value in iobroker. Currently I'm often adding/removing IDs because I'm optimizing my heating. The user mostly does not know which ids he will need upfront. 

Would be great to see that feature in future :)

But great work! Thank you all for that!

fredlcore:
Not sure if I understand this correctly, but could it be that you, @hacki11, are talking about receiving parameter values from BSB-LAN via MQTT and you, @torbennehmer, are talking about checking if BSB-LAN is still online? These seem two different matters to me.

fredlcore:
Polling parameters via MQTT should already be possible via @Luposoft63's code additions.

1coderookie:
> Is the manual already updated with the details?

Updated regarding the new functions polling and setting initiated from the broker side - yes, you can find a little example there:
https://1coderookie.github.io/BSB-LPB-LAN/kap08.html#8212-mqtt
https://1coderookie.github.io/BSB-LPB-LAN_EN/chap08.html#8212-mqtt

hacki11:
If I understand this right, I can use /JW to change which parameters will sent by mqtt, right?

Does this require a reset or just write the config and wait for new values?

1coderookie:
At the BSB-LAN side? I don't know - you're the JSON-API-expert ;)

hacki11:
Hehe yeah I know the API but not the behavior or possibilities :)

1coderookie:
;) But afaik isn't that the point anyway, cuz nothing changed on that side (imho). What's new is the function @Luposoft63  implemented: sending a query and/or a SET-/INF-command from the _brokerside_ towards BSB-LAN - that wasn't possible before.

hacki11:
I could not find the query feature over mqtt in the manual. Just how to write values. That's why I asked about the /jw because I saw no other way :)

1coderookie:
Did you look for that in the PDF? Then I probably didn't uploaded the updated online version. It's always better to look up in the online version. It's definitely in there: "Der Befehl `set mqtt2Server publish BSB-LAN /700` sendet vom MQTT-Broker namens “mqtt2Server” den Befehl “/700” mit dem Topic “BSB-LAN” und bewirkt eine Abfrage von Parameter 700."

fredlcore:
Again, I think you're talking about two different things. @hacki11 wants to change the parameters for MQTT. This can be done either through the /L URL command and probably also through /JW which changes the configuration. Maybe @dukess could say something about the /JW syntax? This could also be relevant for the manual.
I'm not sure about the MQTT features right now. What I implemented in the beginning is that the parameters defined with /L (log_parameters) could be published to a MQTT broker in a given time interval. @Luposoft63 then added functionality to set parameters from the broker to BSB-LAN. I'm not sure if the broker can publish parameters upon request.

hacki11:
Hm i don't think we talk about different things. If the luposoft feature allows asking for new values like the last comment says that's exactly what I'm looking for

fredlcore:
Yes, that is relatively new (a few days), but the discussion before was between @torbennehmer's features, the question of changing the parameters via configuration (which would take effect for the peridically sending of parameters) and the question to query parameters. To me, that's three different things, but as long as you all get the (correct) information you need out of it, I'm fine with it :).

1coderookie:
> I would mention it, as there currently is no magic behind it. Just drop me a quick hint via torben nehmer net (in German or English) and I'll get in touch with a few explanaitions and an example config for Home Assistant.

@torbennehmer : Thanks for your offer, I've sent you an email this morning.

dukess:
/JL format :
```
{
  "0": { //Option output order (as listed in "config[]" table)
    "parameter": 0, //Option ID (as listed in cf_params enum)
    "type": 1, //from cdt_params enum. 0 - byte array, 1 - unsigned byte. can be various format, 2 - TEXT field with format for unsigned integer, 3 - TEXT field with format for unsigned long integer, 4 - TEXT field (string), 5 - TEXT field with format for MAC input/output, 6 - TEXT field with format for IPv4 input/output, 7 - TEXT field with format for programs list input/output, 8 - TEXT field with format for MAX! devices list input/output, 9 - TEXT field with format for DHT bus
    "format": 2, //from cpi_params enum: 0 -not displayed, 1 - HTML text input, 2 - HTML two-state dropdown (or custom switch), 3 - HTML dropdown.
    "category": "CAT_GENERAL_TEXT", // Category name from ccat_params/catalist enums
    "name": "CF_USEEEPROM_TEXT", //Option description
    "value": "150", //Option value
    "possibleValues": [ //Possible values for options with "format" = 2 or 3.
      { "enumValue": "105", "desc": "MENU_TEXT_OFF" },
      { "enumValue": "150", "desc": "MENU_TEXT_ON" }
      ]
    },
....
}
```

/JW format
Request:
```
{
{"parameter": 0, //Option ID (as listed in cf_params enum). "parameter" field in /JL
"value": "150"}, //New value 
{"parameter": 35, 
"value": "2"},
...
}
```

Answer
```
{
  "0": { //Option ID (as listed in cf_params enum). "parameter" field in /JL
    "status": 1 //Status (0 - error, 1 - ok)
  },
  "35": {
    "status": 1
  },
...
}
```

All parameters are passed as strings to simplify and reduce the size of the parser on the BSB-LAN side (the parser is the same for processing parameters when sending the form (BSB-LAN web interface) and JSON)

torbennehmer:
@1coderookie Got the mail, I'll put you something together.

as for the rest: Would it be an idea to move this discussion into the forum :-)

fredlcore:
@dukess: Thanks! @hacki11, I think this structure better fits into the API description rather than the manual?
@torbennehmer: Agree :). Can't close this here, though, since it's a PR and not an issue.

fredlcore:
Thanks!

PR #334, opened 2021-02-08
Fixes and cosmetic changes

dukess:
Fix: address of pointer -> pointer
Static definition of arrays of floats.
ARM -> __arm__
Cosmetic changes: 
custom_float and custom_longs will be enumerate from 1 not 0 in Category 49
"Raumthermostat 3" now use ENUM not SWITCH.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -6918,7 +6918,7 @@ uint8_t pps_offset = 0;
                   pps_values[PPS_E73] = msg[2+pps_offset];
                   break;
                 case 0x69: break;                             // Nächste Schaltzeit
-                case 0x79: 
+                case 0x79:
                 {
                   if (pps_wday_set == false) {
                     pps_values[PPS_DOW] = msg[4+pps_offset];    // Datum (msg[4] Wochentag)
@@ -9287,7 +9287,7 @@ void setup() {
     EEPROM_ready = false;
     SerialOutput->println(F("EEPROM not ready"));
   }
-  #endif
+#endif
 
   pinMode(LED_BUILTIN, OUTPUT);
 
@@ -9476,9 +9476,9 @@ void setup() {
 #endif
   }
 
-  #if defined(__arm__)
+#if defined(__arm__)
   pinMode(temp_bus_pins[0], INPUT); //RX-pin of hardware serial on Due
-  #endif
+#endif
 
   bus = new BSB(temp_bus_pins[0], temp_bus_pins[1]);
   setBusType(); //set BSB/LPB/PPS mode
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -575,7 +575,7 @@ byte programWriteMode = 0; //0 - read only, 1 - write ordinary programs, 2 - wri
 
 ComServer *server;
 ComServer *telnetServer;
-Stream* SerialOutput;
+Stream *SerialOutput;
 
 //BSB bus definitions
 BSB *bus;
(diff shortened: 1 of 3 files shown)


BSB_LAN/BSB_LAN.ino
```diff
@@ -9289,8 +9287,7 @@ void setup() {
     EEPROM_ready = false;
     SerialOutput->println(F("EEPROM not ready"));
   }
-  pinMode(19, INPUT);
-#endif
+  #endif
```
fredlcore, line 9290: Indention?

BSB_LAN/BSB_LAN.ino
```diff
@@ -9478,6 +9475,11 @@ void setup() {
     temp_bus_pins[1] = 18;
 #endif
   }
+
+  #if defined(__arm__)
```
fredlcore, line 9479: Indention?

BSB_LAN/BSB_LAN.ino
```diff
@@ -9478,6 +9475,11 @@ void setup() {
...
+  pinMode(temp_bus_pins[0], INPUT); //RX-pin of hardware serial on Due
+  #endif
```
fredlcore, line 9481: Indention?

fredlcore:
Thanks!

BSB_LAN/BSB_LAN.ino
dukess: In fact, pinMode(19, INPUT) is equivalent to pinMode (temp_bus_pins[0], INPUT) for Due, but allows us not to forget about the pin number when we make further changes.
Test it on my Due without problem.

fredlcore: Sorry, I didn't mean "intention", but "indention" because it looks here that the `#if` is not left-aligned but indented?
Your change from 19 to the variable is really useful, of course!

fredlcore: Actually, I'm not sure how we should deal if the indention of `#if` preprocessor directives. I think they should generally start at the beginning of the line because they have their own hierarchy. But I don't know if we should then start to indent them to the right when nesting starts or not. Maybe it's a good idea to indent them if they are nested, so we better see what depends on what. Could uncrustify do this, too?

BSB_LAN/BSB_LAN.ino
```diff
@@ -4917,16 +4916,15 @@ int set(int line      // the ProgNr of the heater parameter
...
-        int hour=0, minute=0, second=0;
-        strcpy_P(sscanf_buf, PSTR("%d.%d.%d"));
-        sscanf(val, sscanf_buf, &hour, &minute, &second);
-        setTime(hour, minute, second, weekday(), 1, 2018);
-//        printFmtToDebug(PSTR("Setting time to %d:%d:%d\r\n"), hour, minute, second);
-        pps_time_set = true;
-      }
-      case VT_HOUR_MINUTES:
-      {
+      case VT_HOUR_MINUTES: {
```
fredlcore, line 4919: No, these two are different cases, I just missed a `break` as you pointed out correctly.

dukess: I leave comment with warnings and revert my changes.

fredlcore:
As for the changes to the PPS parsing, I understand that we can consolidate some of the `case` segments, but does it really save space in the final code? 
Now, every incoming telegram type has its own `case`, so it's easier to find it later on. Still possible with your changes, but I'd only do it if it really saves significant space.

dukess:
I don't think we'll save space. My changes were only aimed at eliminating ambiguous situations.

fredlcore:
Ok, great, thanks though!

PR #349, opened 2021-03-15
Display grid on chart

dukess:
Just insert piece of code from https://github.com/dukess/bsb_lan_ajax/blob/master/ajax.html


fredlcore:
Thanks!

PR #355, opened 2021-03-20
OpenAPI specification 2.0

hacki11:
This is the API specification.
The link to the UI will be https://editor.swagger.io/?url=https://raw.githubusercontent.com/fredlcore/bsb_lan/master/openapi.yaml when merged.

To review, it is available in via my fork:
https://editor.swagger.io/?url=https://raw.githubusercontent.com/hacki11/bsb_lan/master/openapi.yaml


fredlcore:
Great, thanks a lot!

1coderookie:
Great! So I can mention the API with the given swagger-link in the manual now also?
@hacki11: The name of the repo: ../fredlcore/BSB-LAN/.. - the current (old) "bsb_lan"-version still works, but maybe it would be better if you'd change the link in the future..? Just a suggestion.. ;)

hacki11:
I agree to both points. I didn't get the change. Need to be fixed in Link and yaml.

I will start to make the iobroker adapter compatible to the current version now. If find something I will update the yaml accordingly. Perhaps some questions will be incoming the next days :)

1coderookie:
@hacki11 
I just tested the function /JC={parameterIds} and /JQ={parameterIds} for generating the Curl and URL commands to put it in the manual as an example (we talked about it via email some time ago). 
Therefore I entered two parameters in the field: "700,8300". The commands don't show the comma though, instead it shows `%2C` - so I guess that won't work at the end..?! 
Here's a screenshot:
Bildschirmfoto von 2021-03-21 12-13-25
  
At the "/JQ=.."-field you are giving an example of how to put in the IDs (quote: Example : List [ "700, 710" ]), so I also tried that, but that also doesn't result in the expected command (with comma separated IDs), it even added more stuff:
Bildschirmfoto von 2021-03-21 12-21-33

hacki11:
Thanks for the review!
Regarding the comma I don't think I can do something because it seems to be reserved character. According to the RFC that comma should be encoded and decoded by the Webserver. 

https://stackoverflow.com/a/8828831

Are you decoding uri strings? Then it should not matter if decoded or not.

But I will check if I can find a workaround here.

The [] issue will be fixed by me

1coderookie:
> Are you decoding uri strings? Then it should not matter if decoded or not.

Hmm, not sure what you mean by that..
I just did a quick test and entered url/700%2C8300 (using Firefox) - and it doesn't work, it only shows the 700 within the output of the webinterface.  Dunno if that 'helps' or not though..

1coderookie:
..sorry, I got it wrong previously, now I did the test with url/JQ=700%2C8300 (instead of /JQ=700,8300) - but it's the same though, just the 700 is given back..

hacki11:
Special characters have to be encoded in the url by the client. e.g. "," => %2C
The Webserver has to decode the url string %2C => ",".
There are several characters which normally are encoded.

There would be a workaround, but it will not work because the url is not conform to query params.

Normally a url would look like:

http://bsb/JQ?parameters=700,710,720
If we would have that format (standard query parameters) i could set `allowReserved: true` and it would work. I could even configure a array of integers instead of a string which would improve the clients usability.

But we have non standard here:
http://bsb/JQ=700,710,720
That means, `JQ` is the endpoint and `=700,710,720` ist the path parameter. Path parameters must be encoded on client side and decoded on server side.

I hope i could explain it

hacki11:
> Hmm, not sure what you mean by that..
I meant, is the bsblan webserver decoding incoming urls.

fredlcore:
No, it does not because so far the need to hasn't arisen yet. The browser should send the URL unencoded, AFAIK encoding only takes place in forms etc. That seems to be the case here when you submit the form with the payload, but this is not something that BSB-LAN needs to handle in a normal environment.

1coderookie:
To make it clear: the 'problem' that appeared with the new JSON-API-docu is, that the examples that are built using the interactive 'generator' of the docu don't put out the commas as needed, instead they are replaced by `%2C`. So the URL/Curl which is generated by the swagger-thing is not anymore url/JQ=700,8300 but it is url/JQ=700%2C8300 and so on. I don't know if or how a solution could be found for that, I just noted that behaviour and reported it. 
So if swagger can't do it in the right way and it doesn't make sense to decode the url on the BSB-LAN-side, than we maybe only can point that out as a prominent note within every field of the API-docu or so that users have to replace the %2C from the generated examples manually with a comma to make it work..?

hacki11:
Chrome only encodes "unsafe characters". "," seems not to be unsafe.

https://developers.google.com/maps/documentation/urls/url-encoding
https://de.wikipedia.org/wiki/URL-Encoding

Regarding the RFC the special should be encoded on client but dont' need to be.
Regarding the RFC a webserver needs to support encoded urls, it is not optional.

I do not agree with "swagger can't do it in the right way" because it is conform to the standard and right in the meanings of HTTP.
What we found here is, that the bsb webserver does not respect the RFC regarding encoding/decoding (doing it the right way).

I would say BSB Server is not compatible to swagger because it is missing support of that.
If the "TryOut" of the comma separated endpoints shall be usable, i propose to add url decoding.

1coderookie:
> I do not agree with "swagger can't do it in the right way" because it is conform to the standard and right in the meanings of HTTP.

Yes, you are right, regarding the standards it's correct as you explained it (thanks for that btw) - sorry for my bad expression ;)

> If the "TryOut" of the comma separated endpoints shall be usable, i propose to add url decoding.

Well, now that we have this cool docu and users have the possibility to generate the needed urls, of course it would be nice if they could also just copy&paste them :) But if or not to implement that decoding function isn't up to me, I don't know anything about that stuff..

fredlcore:
/JQ was originally meant only for receiving HTTP POST data. Due to a bug/feature, it also accepts comma separated parameters (as was already the case with /L etc.). And yes, of course Swagger is doing it right and BSB-LAN would have to deal with it if that was necessary, but BSB-LAN is not a full-featured webserver (and never intended to be one). Weighing adding the functionality of adding url-decoding versus the increased size in code, I would vote against adding this functionality if it was only required in this special circumstance.

fredlcore:
What I could offer is to accept both "=" and "?" as a separator between the command and the paramter. If I understood you, @hacki11, correctly above, that would be an option with `allowReserved: true`?

hacki11:
I have added a Swagger Entry with query parameter, i hope you can access here:
https://app.swaggerhub.com/apis/hacki11/BSB-LAN/2.0#/Parameter/listParameters2
(JQ2)

I configured an array of integers and don't even had to work aorund anything. The commas are preserved as needed.
This adds type safety on client side, meaning i can not enter "asdf", but just numbers.

It would be support of `?parameter=` in addition to `=`.

Don't get me wrong here, please. I don't want to drive you adding this RFC stuff to the web server. Thats not my intention. I just wanted to explain what happens why :) Personally, i don't need it to be but perhaps the people who are new to the api and want to play around with it.

fredlcore:
Ok :)
I just tested the link above and entered `700,710,712`, but Swagger colors the field red and shakes it because it seems to accept numbers only?

fredlcore:
Is it possible to change the server? It is `bsb` by default, but the pre-defined hostname is `bsb-lan`(if Swagger is able of accessing the local network at all)

hacki11:
Because it is an array now you have to click addItem. No need to know that is commaseparated anymore.
But it can also be configured to be a string again.
image

The Serveraddress issue, is really annoying - i could not find a solution to change that. 
The reason for that is: Most of the use cases are provider of an existing backend with a known url.
In our case everyone has it's own backend and the server address is not fixed anymore. I will deep dive to see if there is a solution but i fear not.

fredlcore:
Ah, ok, thanks!
As for the server address, if the call is made from the Swagger server, then of course it does not matter. If the call is made from the local browser, it would make a difference to change it in the yaml configuration from bsb to bsb-lan because the latter is now the standard mDNS hostname and will work on most (recently updated) installations.

fredlcore:
Merged your pull request, but the last URL still states "bsb" and the other ones state "bsb-lan", but I don't have the "Add Item" option there?

hacki11:
This is the official documentation link:
https://editor.swagger.io/?url=https://raw.githubusercontent.com/fredlcore/bsb_lan/master/openapi.yaml

The Add-Item in JQ2 was just a proof of concept if bsb-lan webserver would support query parameters. This is nothing what will work now and was just to try if it would work from swagger perspective.

fredlcore:
Ok, understood. Unfortunately, calling `bsb-lan` does not work locally, I always get a network error...

hacki11:
After some research it is not possible to route the requests to localhost. The calls are always made by the swagger-ui backend.
I didn't find a service where that is supported.
This makes the whole Try Out feature completely useless :(

hacki11:
Swagger-UI can be hosted locally with a docker image. I tried to setup a simple docker-compose file to start it quickly but we have another issue.

For stuff like swagger-ui (where the page is not within the same url as the bsb-lan itself), browsers ask the servers if it is allowed (CORS) because otherwise they would leak data to a foreign system. They do this with a HTTP OPTIONS call before any other GET or POST.
Long story short, to get it done, BSB Lan must respond to the HTTP OPTIONS request with a response where cors is allowed from any source.
https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/cors.md
The alternative is, hosting this on bsb-lan itself but i don't think we have resources for such an application :(

fredlcore:
Can you tell me exactly what response would be necessary for the HTTP OPTIONS to work? However, installing a docker instance is already quite a bit of work. Plus, circumventing CORS could pose a security risk, so I'm not sure if it's worth the effort for a nice, but probably niche feature...

hacki11:
I think this explains it well. There are some requirements like Options must be available without auth, cors headers must be in every response and a few more.
https://support.smartbear.com/swaggerhub/docs/ui/cors.html

I agree with you. Modern web solutions and tools meeting embedded systems won't fit nicely because of limited resources. 
I think I will try to get Postman running for that, this is a standalone application which can load openapi spec. Perhaps that works better. Not a click and run solution but as we found out here, swagger isn't either

fredlcore:
Ok, thanks for all your efforts nevertheless, really appreciate it!
What I could think of is to set up a test instance that is connected to one of the heating controllers I have laying around here and which are not connected to a heater. Probably would still make them read-only, but then could make them available on something like http://demo.bsb-lan.de:1984 or so. Then the Swagger approach would still work (unless it gets abused).

hacki11:
You need also cors for that.

Interesting solution I never thought about. Personally, I would rather like to use my own system for checking out the api. I would also not expose the BSB to the public for that because of security.

fredlcore:
Ah, ok, if CORS is still necessary, then that is not an option (and you are right about the security aspect as well, although with an embedded system there is much less risk of a usable exploit compared to a system running on a Raspberry Pi or so)

PR #369, opened 2021-04-29
Merging "readonly" and "writeonly" fields into "readwrite" field, code reduction

dukess:
Unfortunately without testing because my ethernet shield died some days ago.
I will try test these changes with Mega on weekend.


dukess:
Frederik, can you test /Y command? I disabled this command in my MEGA.
Other results looks fine.

Two bugfixes added.

Private question: how to test BSB-Lan board? My Due is blinded: any query return "query failed". LED on board is shining without blinking.

BSB_LAN/BSB_LAN.ino
```diff
@@ -1996,10 +1992,10 @@ void SerialPrintData(byte* msg) {
...
  * *************************************************************** */
 void SerialPrintRAW(byte* msg, byte len) {
-  for (int i=0;i<len;i++) {
-    SerialPrintHex(msg[i]);
-    printFmtToDebug(PSTR(" "));
-  }
+  int outBufLen = strlen(outBuf);
+  bin2hex(outBuf + outBufLen, msg, len, ' ');
+  printToDebug(outBuf + outBufLen);
+  outBuf[outBufLen] = 0;
```
fredlcore, line 1998: This code is less readable than the previous version. Does it have any significant advantage?

dukess (commented):
It allows to save ~512 bytes of flash.

1coderookie:
> Private question: how to test BSB-Lan board? My Due is blinded: any query return "query failed". LED on board is shining without blinking.

If the LED _never_ flashes/blinks when there's traffic on the bus: Did you reconnect the adapter and maybe e.g. used G+ instead of CL+ by accident? 
Other thought: I recently had a similar problem and it turned out that one of the optocouplers was defect.

fredlcore:
Thanks! As for the Due: Do you see incoming bus messages when using the serial monitor? If not, then something's wrong with the RX path. To test the TX path, you could try and set a generic parameter, such as 700 and see if these changes take effect on the heater. If not, then there's (also) a problem with the TX path. With a multimeter, you could test the RX and TX testpoints and see if they show 3.3 Volts. Other than that, it's a bit difficult to run a diagnose from the distance...

dukess:
@1coderookie @fredlcore thanks for advices. I do measure and got values:
TX1 pin - 3.3V
RX1 pin - 0.02 V
VCC testpoint - 2.82V
TX testpoint - ~0V
RX testpoint - ~0V

BC547 С - 15V, B - 0.55V, E - 0.55V
BC557 С - 2.82V, B - 3.3V, E - 0,04V

fredlcore:
Sorry for the delay in response - what makes me wonder is the 2.82V on the VCC testpoint. That should bei 3.3 as with TX1 because it's just an ordinary GPIO. 
You could try and wire it up with DuPont cables and then use the 3V3 pin instead of pin 53.

dukess:
I will make a clarification: 2.82 V with bus connected, 3.3 V without board or bus connection.

fredlcore:
I will have to measure myself the exact values again, but one more thing you could try: Start the Due, wait until the device detection is done (you will get device family 0 and device variant 0) and then try to set a parameter that has only one entry in _defs.h with DEV_ALL and see if that one changes the value on the heater. If yes, then the TX path is working fine and the problem is in the RX path. If it is not working, then we will have to check again.
By the way, has the adapter worked on the Due before at some point or is this the first run?

PR #371, opened 2021-05-04
changing programIsreadOnly() behavior

dukess:
when programWriteMode == 2 we return true when FL_RONLY flag is set


fredlcore:
Thanks!

PR #383, opened 2021-07-20
+ru . de: restored 2 strings in EXTIF section.

dukess:
EXTIF - for external interfaces which can be updated rarely


fredlcore:
Thanks, but what is this for?

dukess:
EXTIF_* strings was introduced by me for my project and not used in BSB_LAN. I can rewrite code for avoiding strings duplicates but little bit later.

fredlcore:
Ok. I have a script that removes string duplicates in the LANG files, so no need for you to write one :)...

PR #391, opened 2021-09-17
Improvements for ESP32 platform, fixes

dukess:
fix for BME280 initialization, 
on/off OTA update on the fly
OneWire 4 ESP32
info about hardware type in JSON
Language resources redefinition warnings elimination (Need somebody for Italian language!!!)
Added support up to 16 BME280 through TCA9548A multiplexor
Disable MQTT when local AP is active


BSB_LAN/BSB_LAN.ino
```diff
@@ -529,62 +529,63 @@ EEPROMClass EEPROM_ESP("nvs", EEPROM_SIZE);
...
+#define TCA9548A_ADDR 0x70
 #endif
 
 bool client_flag = false;
 #ifdef WIFI
   #ifdef ESP32
-#include <WiFi.h>
-bool localAP = false;
-unsigned long localAPtimeout = millis();
+    #include <WiFi.h>
```
fredlcore, line 539: Indention is wrong here. Code indention should be independent of compiler directive indention. So `bool localAP = false;` should be on the same level as `bool client_flag = false` etc.

BSB_LAN/BSB_LAN.ino
```diff
@@ -529,62 +529,63 @@ EEPROMClass EEPROM_ESP("nvs", EEPROM_SIZE);
...
-};
+    //#define ETH_CLK_MODE ETH_CLOCK_GPIO17_OUT
+    //#define ETH_PHY_POWER 12
+    #include <ETH.h>
+
+    class Eth : public ETHClass {
+    public:
+        int maintain(void) const { return 0;} ; // handled internally
+        void begin(uint8_t *mac, IPAddress ip, IPAddress dnsserver, IPAddress gateway, IPAddress subnet) {
+          begin(mac);
```
fredlcore, line 558: Why did you remove the ETHClass::begin here? At least here, it still works even if the ETH_CLK_MODE lines above are commented out because then the defaults will be used.

dukess: Don't panic! :) it is just short call of ETHClass::begin(ETH_PHY_ADDR, ETH_PHY_POWER, ETH_PHY_MDC, ETH_PHY_MDIO, ETH_PHY_TYPE, ETH_CLK_MODE);

BSB_LAN/BSB_LAN.ino
dukess: Now it looks better?

fredlcore: Yes, thanks :)!

PR #399, opened 2021-10-09
ESP32: Watchdog, .yaml, ETag header. Web config: new options and basic/advanced config. /CO command. BMP280

dukess:
Default interval before reset: 120 seconds.
WDT_TIMEOUT definition can be used for interval ajustement.
From https://iotassistant.io/esp32/enable-hardware-watchdog-timer-esp32-arduino-ide/

ETag header support for WEBSERVER on ESP32.
"show_unknown" option added to web config.
One of two wifi reconnection code removed.
Test .yaml for ESP32 (Olimex ESP32 EVB) added.
Print configuration (/CO command) added.
BMP280 sensor support added 
4 new options: RX/TX pins and device family/variant 
Basic/advanced config options for webconfig and JSON config.
Some functions was renamed.
Parser for string parameters was rewritten (bug with strange symbols in text fields may fixed). 
Unneeded writings to EEPROM after termination symbols was eliminated.
Writable parameters in Web and JSON interface:
* JSON: new field "readwrite" added. Values: 0 - read/write, 1 - read only, 2 - write only.
* Web: same as https://github.com/fredlcore/BSB-LAN/pull/411

PR #408, opened 2021-10-28
Improve FR translation

fl42:
Small PR to translate to French params/states that appear often


fl42:
Well... The parameter 701 [should not be queryable](https://1coderookie.github.io/BSB-LPB-LAN_EN/chap06.html#64-simulating-the-presence-function), why do I have values?

fredlcore:
Well,, it could be the state of the push button, but at least for me, this is not the case. Still, it's up to the heater to send or not send a return value.

fl42:
Got it, but that is still strange that the value is "" in JSON and MQTT, but not on the web interface

`http://IP/JQ?query=701` returns
```
{
  "701": {
    "name": "Präsenztaste (temporäre Abwesenheit)",
    "dataType_name": "ENUM",
    "dataType_family": "ENUM",
    "error": 0,
    "value": "",
    "desc": "",
    "dataType": 1,
    "readonly": 0,
    "unit": ""
  }
}
```

While [on the web interface](https://ibb.co/ssvxrb5) there is a value
Is it expected?

fredlcore:
Yes, that's strange, @dukess, do you have an idea why there is a discrepancy?

dukess:
It is normal behavior: i.e. because we not known value of write-only parameters we send empty value. When /JQ command called for read-only or ordinary parameters it return description ("desc" field) of current value but it isn't possible for write-only parameters.
List of possible values will be send when /JC=701 or /JK=11 commands calling.

dukess:
can change behavior:

```
              if (decodedTelegram.value[i] == '1') value+=1<<(7-i);
            }
          } else {
            value = strtod(decodedTelegram.value, NULL);
+            if (decodedTelegram.readwrite == FL_WONLY) value = 65535;
            if ((decodedTelegram.type == VT_ONOFF || decodedTelegram.type == VT_YESNO|| decodedTelegram.type == VT_CLOSEDOPEN || decodedTelegram.type == VT_VOLTAGEONOFF) && value != 0) {
              value = 1;
            }
```

instead

```
              if (decodedTelegram.value[i] == '1') value+=1<<(7-i);
            }
          } else {
            value = strtod(decodedTelegram.value, NULL);
            if ((decodedTelegram.type == VT_ONOFF || decodedTelegram.type == VT_YESNO|| decodedTelegram.type == VT_CLOSEDOPEN || decodedTelegram.type == VT_VOLTAGEONOFF) && value != 0) {
              value = 1;
            }
```

fl42:
Hello @dukess,
I'm sorry but I'm not able to find the part of code to patch, where it is located?

dukess:
Sorry, this patch did after my changes which not accepted yet.
Look for 
if ((decodedTelegram.type == VT_ONOFF || decodedTelegram.type == VT_YESNO|| decodedTelegram.type == VT_CLOSEDOPEN || decodedTelegram.type == VT_VOLTAGEONOFF) && num_pvalstr != 0) {
              num_pvalstr = 1;
and insert 
if (decodedTelegram.readwrite == FL_WONLY) num_pvalstr = 65535;
below.
This will display "---" for write-only ENUM parameters or "on|enable" for switches.

JSON will still return empty value.
May we need one more field to JSON like "writeonly":true/false or something else for API improvement.

fl42:
That makes sense, thanks

fredlcore:
Do we have a read-only flag yet? Otherwise, both would make sense, I guess...

fl42:
@fredlcore Yes readonly flag exists, see [my comment](https://github.com/fredlcore/BSB-LAN/pull/408#issuecomment-955691183)

dukess:
I add new field in JSON: "readwrite". Values: 0 - read/write, 1 - read only, 2 - write only. This field can replace "readonly" field in future.
https://github.com/fredlcore/BSB-LAN/pull/399/commits/2acaf46161efd0fa2d165c1ee8b0e5cea2289266

fredlcore:
I think we should rather replace it now before the (still delayed) release of 2.0, or what do you think, @hacki11?

fredlcore:
Ok, I think we should then proceed and replace the readonly field with the readwrite field. Has this already been included in this PR, @dukess?

dukess:
"readwrite" is already presented in master branch.

fredlcore:
Ok, thanks!

PR #411, opened 2021-11-01
[fix] Set num_pvalstr to 65535 for write-only parameters

fl42:
Following discussion in #408


fredlcore:
@fl42 and @dukess: This PR has conflicts that need to be resolved - can anyone of you have a look at it? Thanks!

dukess:
It can be declined because same fix contains in my merged PR.

fredlcore:
Ok, thanks!

PR #414, opened 2021-11-07
BME Bugfix. DHT library with ESP32 support. Rounding

dukess:
Bugfix: BME_Sensors was limited to 16 at startup.
DHT library with ESP32 support.
CDT_MAC now can print and parse list of MACs instead single MAC in web and json config..
Some printf() was changed from %d to %u for uint16_t types.
Some pieces of code was moved to functions.
More precise rounding. Fix for https://forum.fhem.de/index.php/topic,29762.msg1192996.html#msg1192996


BSB_LAN/BSB_LAN.ino
```diff
@@ -888,7 +888,7 @@ void checkSockStatus()
     SPI.endTransaction();
 
     if ((s == 0x14) || (s == 0x1C)) {
-        if (thisTime - connectTime[i] > 30000UL) {
+        if (thisTime - connectTime[i] > 30000UL || connectTime[i] > thisTime) {
```
dukess, line 891: @fredlcore Can you review fixes for timers when will you have free time?

fredlcore: Sorry, I can't because I'm only using ESP32/Olimex here, and also the whole checkSockStatus never really worked for me, I think it was a suggestion by a user. Why are these changes necessary and why should there be any problems from it?

dukess: When MCU will work more than 49 days millis() will overflow (because it return unsigned long) and starts from 0. And difference between thisTime and connectTime[i] will be always less 30000.
P.S. I will mark other lines with same problem.

BSB_LAN/BSB_LAN.ino
```diff
@@ -4617,9 +4656,10 @@ void loop() {
 
     // Handle PPS MCBA heaters where BSB-LAN has to act as a master and treat the heater as a room unit %-/
     if (pps_values[PPS_QTP] == 0xEA && pps_write == 1) {
-      if (millis() - pps_mcba_timer > 500) {
+      unsigned long milliseconds = millis();
+      if (milliseconds - pps_mcba_timer > 500 || pps_mcba_timer > milliseconds) {
```
dukess, line 4660: millis()

BSB_LAN/BSB_LAN.ino
```diff
@@ -5200,7 +5240,7 @@ void loop() {
           uint8_t found_ids[10] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF };
           if (bus->Send(TYPE_QINF, 0x053D0002, msg, tx_msg, NULL, 0, false)) {
             unsigned long startquery = millis();
-            while (millis() - startquery < 10000) {
+            while (millis() - startquery < 10000 && millis() >= startquery) {
```
dukess, line 5243: millis(). It is unlikely that the /Q command will be called on day 49, but still.

BSB_LAN/BSB_LAN.ino
```diff
@@ -6430,9 +6470,10 @@ void loop() {
 
       mqtt_connect();        //Luposoft, connect to mqtt
       MQTTPubSubClient->loop();    //Luposoft: listen to incoming messages
-
-      if ((((millis() - lastMQTTTime >= (log_interval * 1000)) && log_interval > 0) || log_now > 0) && numLogValues > 0) {
-        lastMQTTTime = millis();
+      unsigned long milliseconds = millis();
+      if(lastMQTTTime > milliseconds) lastMQTTTime = milliseconds;
+      if ((((milliseconds - lastMQTTTime >= (log_interval * 1000)) && log_interval > 0) || log_now > 0) && numLogValues > 0) {
```
dukess, line 6475: millis().

BSB_LAN/BSB_LAN.ino
```diff
@@ -6460,7 +6501,9 @@ void loop() {
   freespace = SD.vol()->freeClusterCount();
 #endif
   if (logCurrentValues && freespace >= MINIMUM_FREE_SPACE_ON_SD) {
-    if (((millis() - lastLogTime >= (log_interval * 1000)) && log_interval > 0) || log_now > 0) {
+    unsigned long milliseconds = millis();
+    if(lastLogTime > milliseconds) lastLogTime = milliseconds;
+    if (((milliseconds - lastLogTime >= (log_interval * 1000)) && log_interval > 0) || log_now > 0) {
```
dukess, line 6506: millis()

BSB_LAN/BSB_LAN.ino
```diff
@@ -6784,8 +6827,8 @@ void loop() {
   }
 #endif
 
-  if (millis() - maintenance_timer > 60000) {
-    maintenance_timer = millis();
+  if (millis() / 60000 != maintenance_timer) {
+    maintenance_timer = millis() / 60000;
```
dukess, line 6831: millis().

BSB_LAN/BSB_LAN.ino
dukess: Yes, it is

> Used here is a unsigned long-variable which is a 4 byte variable.
Maximum value of uint32_t is (2^32)-1.
So when we take another of the same datatype, save 1 to it and deduct 4294967295 it first goes down to zero and then it "overflows" and roll over to its maximum value (((2^32)-1)) which it uses to go down from. So when it is finished deducting it ends up down at 2 :)

So here my mistake. I will revert this change. Thank you.

fredlcore: Great, thanks, so this will apply to the other millis() sections below as well?

dukess: Yep.

fredlcore:
Is it ready for merging? Did you check if your new rounding function also still works for the other data types?

dukess:
I tested it with programs 
720 (positive floats)
750 (percents)
801 (negative floats, temperature)
746 (word, seconds)

fredlcore:
Great, so I can merge this PR now?

dukess:
Yes, sure!

fredlcore ***@***.***> 14 декабря 2021 г. 22:56:21 написал:

fredlcore:
Thanks!

PR #415, opened 2021-11-12
Update LANG_EN.h

GSzabados:
Updated English translation


GSzabados:
Hi @fredlcore and @1coderookie, 

Obviously, you can see that I am not big GitHub user. But #415 and #416 goes together in hand-in-hand.

It is mostly from an LMS14 manual. The codes are usually used on different parameters 8000, 8001, 8002, 8003, 8005, 8008, 8009, 8011. I don't know that I used the right ENUM, but mostly I tried to match the German version.

In some cases I have found some new state numbers, so I tried to add them to the smallest parameter where they happen to be used.

Please have a look and let me know if something is not good.


Gábor

GSzabados:
I've just noticed that there is read.me as well for translation.

`Once you have completed / paused translation and want to test it, set #define LANG in BSB_lan_config.h to your (new) country code. If possible, try to run ./create-undefs.pl (if you have a Perl interpreter) before compilation. It’s not necessary, but prevents compiler warnings. Check if the translations are how you want them to be.`

I cannot do this at the moment as I have some limited resources.

Thanks!

GSzabados:
So, it has turned out that I have actually the Arduino IDE on my laptop, so I could compile the code, found a missing translation, what must have been lost when copied over, and some warning messages pointed me to the fact, that some ENUMs are defined by others.

Like this,

```
BSB_LAN_defs.h:4368:0: warning: "ENUM8010_18_TEXT" redefined
 #define ENUM8010_18_TEXT ENUM8000_17_TEXT
```

or

```
BSB_LAN_defs.h:4360:0: warning: "ENUM8009_12_TEXT" redefined
 #define ENUM8009_12_TEXT ENUM130_01_TEXT
```

I will go along later on the warnings, and update the translations at the right place.

GSzabados:
Merged the other PR here as well, and fixed all the warnings regarding the other definitions.

Meanwhile I have found this in the BSB_LAN_defs.h

```
#define ENUM8008_16_TEXT ENUM8005_16_TEXT
#define ENUM8008_17_TEXT ENUM8000_16_TEXT
#define ENUM8008_18_TEXT ENUM8000_17_TEXT
```

This is a bit weird to assign to state number 17 the value of 16, and to 18 the value of 17.

Even more weird is this:

`#define ENUM8004_95_TEXT ENUM8000_16_TEXT`

This should be "Schutzbetrieb Kühlen" (149)

Meanwhile 

16: Minimalbegrenzung aktiv (22)
17: Anlagefrostschutz aktiv (23)
18: Frostschutz aktiv (24)

It might have been misread from the manual before:

image

As I understand one state/status number should have only a single equivalent text representation for parameters 8000 to 8011. So only just a single set of status number should be defined from 1 to 255 for all the 8000 to 8011 parameters.

fredlcore:
Hi,

thanks for making the effort to add the translations, but adding translations is quite complex because it also involves accurate checking if the same piece of text already exists somewhere eles. If yes, this piece of text is reused/referenced and thus saves flash memory instead of having the same piece of text three times. This is what you described as "weird" ;). If you have translations to add or correct, please open up an issue and describe them there and we'll take care of the rest.

GSzabados:
@fredlcore,

The PR was ready with all the possible options for the 8000 menus.

I cross checked the references and it is added only one place.

What I described "weird", is referencing the wrong text labels.
I wanted to confirm with you that was that intentional or just a typo as added the picture from the manual as well.

I will open then an issue and assign the PR to that one.

Meanwhile could you just have another look on those numbers and texts what I described "weird"?

Best Regards,

Gábor Szabados

1coderookie:
Hi Gábor,
thanks for your effort. I guess you're talking about the LMS14 manual I sent you, right? I also checked other manuals for the parameters that aren't listed in the LMS14 manual (like the 8004 for example). So I checked these ones 
```
#define ENUM8008_17_TEXT ENUM8000_16_TEXT
#define ENUM8008_18_TEXT ENUM8000_17_TEXT
```
within that manual and it seems that you're right, as 
_17 (=23) is "Frost protection plant active" ("Anlagefrostschutz aktiv") and
_18 (=24) is "Frost protection active" ("Frostschutz aktiv")
for both 8000 & 8008, so up to me here's an error (@fredlcore please verify that too).

8000_16 (=22) I didn't find anywhere yet, but it should be the same as  
```#define ENUM8008_16_TEXT ENUM8005_16_TEXT```
where both 8008_16 and 8005_16 (=22) are "Minimum limitation active" ("Minimalbegrenzung aktiv").

8004_95:
As I dind't find 8000_16 (=22), I don't know if this
```#define ENUM8004_95_TEXT ENUM8000_16_TEXT```
is correct or not, but 8004_95 (=149) is "Schutzbetrieb Kühlen". If we take _16 (=22) given as "Minimum limitation active" for all of the 8000-8029, then you're right, this would be an error here also.

I don't understand this change you made though:
Bildschirmfoto von 2021-11-12 08-43-41
8a (=138) is correct, as it is "cooling mode off" ("Kühlbetrieb aus"), but 89 (=137) which you added as "heating mode" isn't even existent in 8004..? So why did you delete _8a and put in _89 here? Or did I miss something here..?

**Comment by fredlcore** [comment]
Please explain to me exactly where you see the error here:
```
#define ENUM8008_17_TEXT ENUM8000_16_TEXT
#define ENUM8008_18_TEXT ENUM8000_17_TEXT
```

1coderookie:
The text for ENUM8008_17 is taken from ENUM8000_16 - right? 
If so, then this must be the same - judged by the description of the manual for option 23&24 at 8000 and 8008: 
> (...) as _17 (=23) is "Frost protection plant active" ("Anlagefrostschutz aktiv") and _18 (=24) is "Frost protection active" ("Frostschutz aktiv") for both 8000 & 8008, (...)

Screenshot for 8000:
Bildschirmfoto von 2021-11-12 09-48-09

Screenshot for 8008:
Bildschirmfoto von 2021-11-12 09-49-24

fredlcore:
Yes, 8008/17 and 8000/16 are the same (text). Is there a difference between "Frostschutz aktiv" and "Anlagenfrostschutz aktiv"? If yes, then there is an error. If it's just a slightly different expression that is for example used interchangeably in different heaters, then it's not an error.

GSzabados:
@1coderookie, yes, I dug through the LMS14 manual and I have found the Brötje ISR-PLUS manual as well.

https://polo.broetje.de/pdf/systemhandbuch_isr.pdf

It seems to be there are state numbers for the 8000-8011 parameters. All defined from 1 to 255. They seem to be the same. Reused all over the place. The manuals suggest that, only some parameters are used by 8000 and not 8009 etc. (Page 184-185)

Screenshot_20211112-035438_Drive

So state number 17 (23) means the same for all parameters: Frost protection plant active

It is enough to define them once for all of them and then it can be referenced.


> So why did you delete _8a and put in _89 here? Or did I miss something here..?

It was cross reference and a typo together. 8011_89 is Heating mode, but I really lost with the cross referencing and the different parameters.

I built a table with the German translation, state numbers, and English translation for state numbers. It was not easy to match things together where parameter numbers and state numbers are all over the place.

Anyhow, the best solution would be to have a library ENUM800X_01 to ENUM800X_ff where all the states are defined for the possible 255 and if 8003_17 or 8009_17 needs a value then all should be referenced to 800X_17.

Please excuse me if I wasn't clear with what I meant to write. I have started this project a bit late on the day and it is even later now.

1coderookie:
@fredlcore : It seems to be a difference _somewhere_ in the meaning, as it's stated different in the different user levels (with different numbers also): left column = enduser (info level) vs commisioning/heating engineer in the right column. Another example:
Bildschirmfoto von 2021-11-12 10-03-51
The heating engineer level seems to be more detailed and offers more descriptions about certain stages, so we should use these descriptions consistent imho.

@GSzabados No no, thanks for your effort, we didn't see this one earlier. Please fix the _8a/_89 mistake though.

parsley:
Usually „frost protection“ is only the boiler itself. „Anlagen Frostschutz“ includes pumps, valves and so on, that are controlled by the boiler but not build into the boiler itself. E.g. external heat circuit pumps. 

So, yes, there is a difference.

fredlcore:
Ok, thanks!

1coderookie:
@parsley That makes sense - thanks!

fredlcore:
@GSzabados: As for 800X as a master reference, we already do that. If a parameter has (exactly) the same options and option values compared to any other, this set of options is referenced to it in _defs.h.

PR #426, opened 2021-12-13
SMD PCB added

MrAntonka:
added smd pcb


dukess:
👍  Board will be more cool if it has more pins out from MCU not OneWire and DHT only.

fredlcore:
Thanks, but as I explained earlier, adding a PCB to the repo will raise questions as to whether this is the same that we're providing, and if people were to rebuild it, they would come up to me for questions mainly. For these and other reasons I would respectfully decline your PR and hope for your kind understanding.

PR #427, opened 2021-12-19
Update Hungarian localization

olympia:
Hi Frederik, hello Ulf,

This is a major update to the Hungarian localization as promised.

Cheers,
Bence


fredlcore:
Thanks!

PR #430, opened 2021-12-28
Added Brötje WGB EVO 20i and WGB ProEVO 15C

jalbersol:
Hi,
I added the family and variant IDs for the following heaters:
- Brötje WGB EVO 20i: 193, 5 (our current heater)
- Brötje EcoTherm Plus WGB Pro EVO 15 C: 98, 100 (our old one)

Greetings
J.

### DIFF
--- a/BSB_LAN/BSB_LAN_config.h
+++ b/BSB_LAN/BSB_LAN_config.h
@@ -1,0 +1,1 @@
+../../BSB_LAN_config.h


fredlcore:
The file listed here is BSB_LAN_config.h which would be the wrong place to do so.
The better way is to send me the output of "Check for new parameters" (URL command /Q) so that I can add it properly and in the process also add any new commands (which may take a bit of time due to my long todo list).

1coderookie:
..and the WGB EVO 20i is already listed in our compatibility list, so it already should be integrated btw..

fredlcore:
Interestingly, we don't have device family 193 in our list yet, so that seems to be new (previous WGB EVO 20 H was device family 163).

jalbersol:
Oops, sorry. That was the wrong file. Should actually contain the file BSB_LAN_defs.h.
+ #define DEV_098_100  98,100 // Brötje EcoTherm Plus WGB Pro EVO 15 C (LMU75.001A100)
+ #define DEV_163_005  163,5   // Brötje WGB EVO 20 i (LMS15)

1coderookie:
..I just wanted to ask where you got the 193 from @fredlcore, cuz ALL the WGB EVO (also the latest i-type) we have listed are family 163, so I assumed that maybe was a typo - now @jabersol sent the 163, so that's ok now. 
@jabersol: As freetz already said, please just send him /Q from every unit together with the corresponding name of that heater, then he'll take care about it (which includes a check for new parameters)..

jalbersol:
Gosh, and a typo too. Okay, maybe it wasn't my day. I will send the result of /Q.

PR #431, opened 2021-12-30
Dynamic URL changing

dukess:


fredlcore:
Great idea, thanks a lot!

1coderookie:
Sorry guys, I don't get it - what is this function for? How & where should I mention it within the manual?

fredlcore:
No need to mention it in the manual, it just automatically updates the link to the sensors in the menu to refer to the correct category number, so that each time (hopefully no more ;)!) a new category is added, the sensor category is updated automatically in the menu.

1coderookie:
Ah ok - well that's cool though! :)

fredlcore:
Especially since I always forget that I need to change something there as well ;)...

PR #433, opened 2022-01-06
Update to HU translations 

olympia:
(somehow has got broken due to recent changes to category strings)

Cheers,
Bence


fredlcore:
Thanks! Yes, I hope that was the last time we had to add new categories ;)...

PR #437, opened 2022-01-15
Fix: Categories with N>63 will not shown in JSON mode.

dukess:


fredlcore:
Thanks!

dukess:
By the way: averages.txt appended every minute and not overwritten. It is 
normal? I found seek(0) but it looks like it not working. mcu - esp32.


fredlcore ***@***.***> 16 января 2022 г. 12:50:26 написал:

fredlcore:
Good call, the problem is that in (currently) line 6632, FILE_WRITE gets redefined to FILE_APPEND. There was some problem with FILE_WRITE and FILE_APPEND on ESP32 when using it with SD card which is why I wrote a workaround for it (I think FILE_WRITE was not available in that way, but FILE_APPEND was). But of course that now also badly affects averages.txt. We'll have to see if this problem still persists with current SDK and if not, then revert back to original FILE_WRITE vs. FILE_APPEND.

fredlcore:
I think I remember now: FILE_APPEND does not exist for Arduino, FILE_WRITE is defined in ArduinoFiles.h as
```
/** Arduino SD.h style flag for open at EOF for read/write with create. */
#define FILE_WRITE (O_RDWR | O_CREAT | O_AT_END)
```
So what I did now in the Arduino section is 
`#define FILE_APPEND FILE_WRITE`
because it seems that FILE_WRITE on Arduino will always append (and create if not existing). 

In our code that should work well because whenever we use FILE_WRITE, the corresponding file has been deleted before. But this is not the case with averages.txt. So far, we assume that FILE_WRITE will remove the file - which it now does with my changes. But not on Arduino, so I guess the behaviour now is reverted. I currently don't have a Due with SD card running here, so maybe one of you guys could test this behaviour? Thanks!

fredlcore:
If now the Arduino is appending averages.txt instead of the ESP32, then we'd have to do a dataFile.remove() before writing the averages. But maybe the seek(0) is working on the Arduino, and FILE_WRITE will take care of the problem on the ESP32, then we're good to go...

fredlcore:
Ok, I just checked the implementation of `seek()` in ESP32 and Arduino, and the former (in FS.cpp) really only seems to return the current position - it does not change the write position within the file, all it does is this:
```
bool File::seek(uint32_t pos, SeekMode mode)
{
    if (!*this) {
        return false;
    }

    return _p->seek(pos, mode);
}
```
No calculations or positioning such as in the Arduino library. 

I assume it will all work now with FILE_WRITE but at the cost that we're recreating the file every minute instead of just overwriting it from the beginning. It shouldn't cost any noticeable processor time, but I'm not sure if there is any extra overhead for recreating the file on the SD card and if this will put an extra/unecessary strain on the write-cycles of the card...

dukess:
Hmmm... seek() implementation exist both for SdFat and ESP32. 
ESP32: seek(0, SeekSet); will to set cursor to begin of file and seek(0, SeekEnd); - to end of file.
SdFat: seek(0); or seekSet(0) will to set cursor at start. seekEnd(0) -at end.

May here is bug in ESP32?

fredlcore:
Maybe I don't get it, but what exactly is called on the ESP32 when it `_p->seek(pos, mode)` is called? I couldn't figure out what the value passed as second parameter to seek() actually does. In the Arduino code you can actually see several calculations regarding EOF etc., but I couldn't find it in the ESP32 code, so I thought they just didn't implement it?

PR #449, opened 2022-03-24
code readability, reducing the number of sensors programs setup points

dukess:
Oh, one more thing: functions tables sizes was corrected for Mega users.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -1205,22 +1205,22 @@ void printcantalloc(void) {
  *   none
  * *************************************************************** */
 uint8_t recognizeVirtualFunctionGroup(uint16_t nr) {
-  if (nr >= 20000 && nr < 20007) { return 1;}
+  if (nr >= BSP_INTERNAL && nr < BSP_INTERNAL+7) { return 1;}
 #ifdef AVERAGES
-  else if (nr >= 20050 && nr < 20050 + numAverages) {return 2;} //20050 - 20099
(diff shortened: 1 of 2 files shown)


fredlcore:
What do you mean that functions tables sizes were correct? The cmdtbl sizes? I think most of them are more than 32767 bytes now, isn't it? 32k is the maximum on the Mega...

dukess:
Yep. I tried to compile for Mega and program was built normally. Sure, they 
size was bigger than flash size.

fredlcore ***@***.***> 25 марта 2022 г. 0:23:08 написал:

fredlcore:
Well, it didn't produce an error before, but there was a memory overflow which resulted to mixed up texts etc. which Mega users have been reporting for a while now...

dukess:
Unfortunately we can't help with this.  :(
I have one more idea: I can add a check for the absence of overlapping number ranges of user sensor programs. This can be useful when users will change the number of sensors.
Or is it not worth thinking about it yet?

fredlcore:
No, no need to help with that - Mega is now hopelessly too small, in many ways, so we just keep it as it is. We could rather think of removing the Mega code to make the code more readable.

As for the overlapping number ranges, I'm not sure I understand what you mean with that? The user parameter numbers are fixed, so how should they overlap?

dukess:
Just an example: if user will set 'Averages" array size more than 50 elements and query programs 20100 and above (DHT sensors), he can get unexpected answer (average value(s)) instead DHT descriptions and values.

fredlcore:
Ah, ok, got it - hm, I would wait for that until we really see the need for it - otherwise we encourage people to make these kind of changes and everybody does their own kind of stuff which makes debugging more difficult...

dukess:
Got you. :)

PR #451, opened 2022-04-04
small fix: compilation will break when averages is disabled

dukess:


fredlcore:
Great, thanks a lot, didn't have time to look into this issue!

PR #453, opened 2022-04-26
Summer time for Baxi Luna Platinum+

dukess:
Correct command codes for programs numbers 5 and 6 for Baxi Luna Platinum+ (Should be used via INF).
P.S. I've never paid attention to it, but today I decided to check it out, and it turned out that it needs to be fixed.


fredlcore:
Thanks!

Issue #452, opened 2022-04-12
[FEATURE REQUEST] MQTT port


tobyDoh:
**Is your feature request related to a problem? Please describe.**
MQTT Port is currently hard-coded in Programm

**Describe the solution you'd like**
Have the MQTT Port avaiable as Option in header as variable or better as option in webinterface

**Describe alternatives you've considered**
changing the coding prior to compiling process

**Describe your own contribution**
i will test the changes

fredlcore:
Since the webinterface is already quite "full", I'd prefer only to add necessary configuration options. What is the benefit of adding a configuration option to configure the MQTT port? Especially in relation to all other network ports we are using? If we start making one of them configurable, we would have to do the same with all the others as well. 
If you are looking for security through obfuscation, it won't bring much in a local network where local users connected to the network can easily monitor the unencrypted communication on all ports.

fredlcore:
Ok, thanks.

PR #457, opened 2022-06-03
Bugfix: EEPROM was overwritten too frequently. Automatic room temperature measurement and sending to boiler in PPS mode

dukess:
EEPROM bugfix! pps_values without FL_EEPROM flag write to EEPROM.
When device working in PPS mode, user can set up to 5 sensors which will measure temperature and send it to boiler (if pps_write mode enabled)
RGT_EMULATOR module should be enabled.
MQTT port now can be configured through web interface. https://github.com/fredlcore/BSB-LAN/issues/452


fredlcore:
Thanks!
Can you explain which PPS values were written so frequently? I thought that we already had a flag which prevented writing to EEPROM?
And what is the use of the five sensors that are transmitted to the heater? Usually this is done through the home automation system which - if required - collects the various sensors and sends a weighted average. Or is it the same sensors we already are offer to the user?
And is it correct that MQTT is now enabled by default in BSB_LAN_config.h.default?

dukess:
> Can you explain which PPS values were written so frequently? I thought that we already had a flag which prevented writing to EEPROM?

Yes, sure: 
when we check if ((flags & FL_EEPROM) == FL_EEPROM && EEPROM_ready), new PPS value already set in pps_values array. So this check will prevent instant overwriting EEPROM cell until changing parameter with FL_EEPROM flag. But when program will change cell with FL_EEPROM flag it will update EEPROM for whole array. So i added additional check for excluding frequently changed cells in writeToEEPROM(). In other words these cells will never to write to EEPROM.

>And what is the use of the five sensors that are transmitted to the heater? Usually this is done through the home automation system which - if required - collects the various sensors and sends a weighted average. Or is it the same sensors we already are offer to the user?

This functionality was added a year ago when we made a room device emulator: the user can specify from one to five programs with temperature values (Dallas, BME280, DHT, or custom_floats) and these values will be averaged and average value will be sent to the heater.
Now, if, in PPS bus mode, the program number(s) is specified in the setting of the sensor list for RGT1, then the temperature value will be set for the PPS_RTI cell. (I hope I understood its meaning correctly)

>And is it correct that MQTT is now enabled by default in BSB_LAN_config.h.default?

No. SD card logging only will be enabbled by default. For MQTT enabling LoggingMode should have 00001000.

dukess:
NOTE: this EEPROM scheme is incompatilbe with old. Users should save config before upgrading nd reconfigure devices after that.

fredlcore:
Thanks for the note - is there any way we can warn people? Or what exactly would happen if you install it over the old EEPROM scheme? I'm just thinking that it would mean that also the device could no longer be accessible if the network configuration is lost?

dukess:
I fix this issue, but we can forget about space saving :)

PR #458, opened 2022-06-11
Adding datalog.txt viewer (*.html in scripts directory), proposal to change /DG implementation accordingly (D3 -> JSCharting)

DE-cr:


DE-cr:
Proposal to changed /DG data plot implementation from using D3 to using JSCharting:

Pros:
+ added zoom and pan capability
+ user can interactively disable plot lines for improved overview and scaling
+ user can interactively highlight plot lines for improved overview
+ saves over 6 KB in compiled BSB-LAN sketch

Tests:
o tested with datalog.txt sizes up to 1.2 MB / 18641 lines
o tested using Chrome@Android+Ubuntu, Edge@Windows11, Firefox@Ubuntu,
  Safari@MacOS, Silk@Android
o tested using window sizes from ridiculously small ;) to UHD
o tested using display rotation on Android
o performance NOT measured and compared to D3 implementation

Cons:
- client loads 768 KB javascript library instead of 152 KB (for D3)
- developer needs JSCharting licence (which is free for non-commercial use)
- with zooming enabled, there seems to be no way to scan plot values
  (tooltip display) on touch screens (no mousemove events)

dukess:
Hi! Can You place here one or two screenshots?

fredlcore:
Yes, screenshots or a live demo of how this would work in BSB-LAN would be great. After reading their website, I still don't understand their licensing model. If it's necessary to register with them for a free license, such license information would have to be transmitted from BSB-LAN to JSCharting. I'm not sure this will work if these requests come from IP addresses from all over the world (well, not _all_ over the world ;) ) because it would look like as if someone (or many people) would have "stolen" the licensing information. I assume this is more geared towards a single website which uses this license in one place (similar to how Google Maps licensing works).
If you can clarify that this is not a problem, I'm willing to look into it.

fredlcore:
Also, what's the file BSB-LAN_datalog-viewer.html file for? I don't find it referenced in any other file?

DE-cr:
As I find the licensing info I can see on JSCharting's website somewhat confusing, I've just sent an e-mail to info@cwshost.com, asking for clarification regarding free licenses. So you may want to hold your horses regarding this pull request from me. Sorry if it should turn out to be in vain!

The *.html in the script directory does what its name says: It allows you to view datalog.txt files provided by /D in BSB-LAN. As
a) I've disabled logging on my BSB-LAN system again and
b) I see the advantage of my JSCharting implementation proposal over the existing D3 implementation in BSB-LAN mainly in the interactive use:
Please try the *.html (which does basically the same thing as the code for BSB-LAN, except for the BSB-LAN header and width restriction, which it replaces by a file upload functionality). Just open the *.html in a browser, then load any datalog.txt file you have on your system (or should I provide a sample?). You can then highlight plot lines (mouse hover over plot line name at the top of the screen), disable plot lines (click on plot line name), click-and-drag the graph to zoom in, pan (or zoom further) once zoomed in (watch the mouse cursor to differentiate between pan and zoom), ...

DE-cr:
I haven't received a reply from info@cwshost.com, yet, but I've managed to find a EULA for JSCharting hidden somewhere, which includes:

"JSCharting may not be used in any other application outside of a web server environment." So scripts/BSB-LAN_datalog-viewer.html is out.

"JSCharting may not be integrated with or incorporated into any software product or script sold or offered to any 3rd party without a separately executed licensing agreement for SaaS or OEM." Looks like including it in BSB-LAN is also not permitted - although the wording here is less clear to me: Does "integrate or incorporate into" mean the same as "use", i.e. does it include SRCing https://code.jscharting.com/latest/jscharting.js (vs. using a local copy)?

I'm sorry for having wasted your time. However, if one day you should be looking for ways to reduce BSB-LAN's footprint, I see about 6 KB of potential in replacing the D3 implementation of /DG with using one of the many  other javascript plot libraries, many of which have licenses that would allow their inclusion in BSB-LAN.

fredlcore:
Thanks for your research, and yes, to me that reads that this definitively prohibits the use of JSCharting in BSB-LAN. But no time was wasted other than yours, so thank you for your effort! I'm certainly open for better ways of charting the stored data, especially if it can save several kB (which seriously is a lot for embedded systems)...

PR #460, opened 2022-06-24
Update LANG_DE.h

DE-cr:
fixed MENU_TEXT_DTR wording


DE-cr:
p.s. now also sent as a proposal (which I've just become aware of)

PR #461, opened 2022-06-25
add id 8739

inuites:
Parameter 8739 is missing from BSB-LAN parameters list.
id 8739 => Ambient relative humidity 1 (English)
I add this new parameter with all corresponding language translations


1coderookie:
Interesting - which device / room unit measures the humidity though?

inuites:
I forgot to specify but I own a QAA74.614/101 which has temperature (id=8740) and humidity (id=8739) sensors. 
That's why the parameter 8739 isn't avalaible if you only have a controller connected to the BSB-LAN.

1coderookie:
Nice, thanks for that information, I never stumbled upon the fact that the QAA74(.611) can also measure humidity! Will mention it within the manual also..
So then ambient relative humidity 2&3 will/must be available as 8769 & 8799 also (I assume) but most likely with a slightly different CoID probably..?

fredlcore:
Thanks from me, too! Did you send me the output of /Q already together with the exact model of your heater?

fredlcore:
Hi @inuites, I'm not sure if you are already using version 3.x of BSB-LAN, but if not, could you maybe try and upgrade your version? You will have to send me a dump to get the full parameter list from me afterwards. But after that, I would like to ask you to do a serial monitor log to see which the telegrams are exchanged between the QAA74 and the heater to identify the room humidity telegram. This is the one we don't know yet and we could add it to BSB-LAN as well so that users could emulate the room humidity sensor by sending values directly to the heater.
Would be great if you could help us with that!

PR #464, opened 2022-06-30
/DG: automatically scale plot to fill whole browser window

DE-cr:
I've eventually figured out how to do this. No more wasted screen estate, no more cut off graphs!

With this implementation, invoking BSB-LAN's /GD page at first shows the graph cut off at the bottom, but that is easily "fixed" by hitting page down once. Then the graph fills the complete browser window (and BSB-LAN's menu is scrolled out of view at the top).


DE-cr:
Both examples above show the initial/top part of the /DG page first, then the page scrolled down for the full graph.
(The data displayed isn't very interesting, but that's not the point here. ;)

PR #466, opened 2022-07-21
feat: added enum-texts for 832, 5890, 5930 and additional entry 616 for FUJITSU 170

Duesentrieb5413:
Added enum-texts for 832, 5890, 5930 
Added additional entry 616 for FUJITSU 170


1coderookie:
Thanks! 
@fredlcore : I wonder if/why we don't have them yet, because they're part of the RVSx1 controller already (and e.g. Trinkwasserfühler B3 should already be existent anyway I would have assumed)..?

Anyway @Duesentrieb5413 - could you please change the abbreviation "Quellenaust'fühler" to the full expression "Quellenaustrittfühler"? 

And just in case you're willing to edit LANG_EN also, here are the belonging translations:
```
"Kondensatorpumpe Q9" = "Condenser pump Q9"
"Trinkwasserfühler B3" = "DHW sensor B3"
"WP Vorlauffühler B21" = "HP flow sensor B21"
"WP Rücklauffühler B71" = "HP return sensor B71"
"Heissgasfühler B81" = "Hot-gas sensor B81"
"Aussentemperaturfühler B9" = "Outside temperature B9"
"Quelleneintrittfühler B91" = "Source inlet temperature B91"
"Quellenaustrittfühler B92/B84" = "Source outlet temperature B92/B84"
"Raumfühler B5" = "Room sensor B5"
```

fredlcore:
@1coderookie Yes, we do have these parameters already, but apparently not the specific option values. And yes, some of the translations already exist.
@Duesentrieb5413 We only add text descriptions (whether it's parameter names or options) in LANG_XX.h if the same phrase has never appeared before. In this case, for example, "Trinkwasserfühler B3" already exists in this line:
`#define ENUM4133_01_TEXT "Trinkwasserfühler B3"`
So in order to save flash space and work for translators, this will be cross-referenced in BSB_LAN_defs.h like this
`#define ENUM5930_13_TEXT ENUM4133_01_TEXT`
and so on.

I guess we should mention somewhere that if people want to contribute translations (which we of course really welcome), they should get in touch with us first to understand the mechanism that the language files are designed upon in order to save everyone unnecessary work...

Duesentrieb5413:
Removed duplicate texts from German language file and added translations to English language file. Furthermore pushed automatically adapted web files to be contained in the pull request.

fredlcore:
Mit der Version 3.0 und den reglerspezifischen BSB_LAN_custom_defs.h wird sich die Übersetzung der Parameter erübrigen, trotzdem vielen Dank für Deinen Einsatz!

Duesentrieb5413:
@fredlcore : why is this closed without adding the provided parameters/texts to the code?
Several new texts were added, which I currently cannot find in the appropriate BSB_LAN_custom_defs.h, e.g. 
- ENUM5890_24_TEXT 
- ENUM5890_25_TEXT 
- ENUM5930_11_TEXT
- ENUM5930_13_TEXT
- ENUM5930_14_TEXT
- ENUM5930_15_TEXT
- ENUM5930_16_TEXT
- ENUM5930_17_TEXT
- ENUM5930_18_TEXT
- ENUM5930_19_TEXT
- ENUM5930_1a_TEXT 
- ...
Or were they added in another file that I am currently not aware of?

fredlcore:
See here: https://github.com/fredlcore/BSB-LAN/discussions/482
Translations will be taken from the heater's definitions. The current BSB_LAN_custom_defs.h is still reflecting the old "complete" list, so they're not included there. But since this is legacy, it won't make sense to include new translations.

Issue #432, opened 2021-12-31
[FEATURE REQUEST] Add UDP broadcast of log parameters


fredlcore:
A year ago or so, we had two users who implemented a UDP transmission of parameters to use with their DIY room units to display various parameters there.
I was thinking whether it would be possible to use our current MQTT functionality and just add a few command to send the same payload via UDP, probably just as a broadcast to keep it simple. The receiver(s) could then parse the payload and display whatever they need. 
A receiving function could also be implemented the same way so that the MQTT processing routines could be used, like for setting a parameter (DHW push after pressing a button on the room unit etc.).

DE-cr:
Add one more user who did this just recently (https://github.com/DE-cr/BSBmonCR). :)

I'd be willing to do an implementation proposal - once some design decisions have been made:
1. Shall the parameter list and send interval be shared with logging (less implementation effort) or shall a separate configuration set be implemented (my preference)?
2. What data format shall be sent? The same as in logging (e.g. "364593010;01.05.2022 00:00:15;8314;Kesselrücklauftemperatur Ist;66.7;°C"), or a more concise one (e.g. BSBmonCR would use "8314:667" for the same data)?
3. Or should UDP sending be more like MQTT, as fredlcore suggested? I haven't looked at MQTT, yet, but would propose to keep the UDP data as easily parseable (and concise?) as possible.

Personally, I would
a. definitely use UDP broadcasting, not sending to one (or multiple?) dedicated receivers
b. not allow parameter setting via UDP. This can already be easily done via a simple "HTTP GET /S<x>=<y!z>", and UDP's broadcast capabilities are not required here, but a confirmation would be nice.

@fredlcore, let me know what you think.

fredlcore:
For now, I would keep the UDP logging parameters the same as the general logging parameters. I don't think someone will be using SD-card logging and UDP logging **and** use different parameters on both.
The issue I see here is the update interval. If you were to use UDP for updating, say, in a project like the one you mentioned, it makes a difference if you poll 20 parameters within 10 seconds (which will clog the bus and BSB-LAN) or just two or three. So I'd suggest adding a UDP update interval in the configuration.
As for the data format, I think it's easiest to take the format we already use for logging on SD-card or MQTT. That way, the remote device does not need to make a look-up to find out what the parameter number means, what unit it is in etc. and transmission time should not be affected in a noticeable way.

DE-cr:
[...]

Issue #455, opened 2022-05-24
[FEATURE REQUEST] Report RSSI


designer2k2:
**Is your feature request related to a problem? Please describe.**
I would like to know the RSSI (Wifi Signal Strength) over time, basically to see if the signal is constant or maybe fluctuating during the day.

**Describe the solution you'd like**
Getting it via MQTT or being able to query it like any other parameter. 
Maybe like the OneWire that start at 20000 this "internal" information could start at 30000?

The ESPhome devices have a function like this: https://esphome.io/components/sensor/wifi_signal.html

**Describe alternatives you've considered**
If i could see the RSSI on the setting page (/C) maybe below the uptime this would also work, but not as good.

**Describe your own contribution**
I could help to test it.  But for more i would need a hint on how to implement it as im quite lost in the code.

The Serial debugging gives the RSSI: https://github.com/fredlcore/BSB-LAN/blob/f83c20b8c962adb68e1b7cfd3fcda3f7bef9ac76/BSB_LAN/BSB_LAN.ino#L6925

**Additional context**
This only makes sense on WIFI enabled devices.

DE-cr:
@designer2k2, I see how this could be useful.
Have you checked whether you could easily receive this information from your WLAN router instead?
Displaying the RSSI on the /C page should be really simple to implement. It that would be good enough for you, let me know if I should do an implementation proposal.
Creating a fake BSB parameter would be more work, and might require some design decisions which I'd rather leave up to @fredlcore.

fredlcore:
I'm relatively strict when it comes to adding new features due to the limited amount of flash memory we have on the Due and eventually also on the ESP32 (which has more Flash, but half of it is reserved to perform OTA updates and the code in general is much larger than for the Due due to the more extensive framework), and especially when it comes to features that are not directly connected with the core functionality. 
The RSSI values are usually only needed once when you setup the device to figure out a suitable place, so a permanent function for that would not be necessary. Additionally, you can get this value usually from your router, and if not, you can add the suggested code to BSB_LAN_custom.h and send it to the debug output, that's what the custom code function is designed for.

PR #471, opened 2022-08-18
Optional logging via UDP broadcast added (configurable, same parameters and format as in SD card logging)

DE-cr:
c.f. https://github.com/fredlcore/BSB-LAN/issues/432

tested:
- logging to: none, SD, UDP, SD+UDP
- on ESP32, via WiFi

not tested:
- logging to: MQTT, averages (don't know how to receive/interpret)
- on Arduino, via LAN (not available)

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -6585,13 +6585,14 @@ void loop() {
           }
         }
       }
+      IPAddress broadcast_ip;
       if (LoggingMode & CF_LOGMODE_UDP) {
         #ifdef WIFI
           IPAddress local_ip = WiFi.localIP();
         #else
           IPAddress local_ip = Ethernet.localIP();
         #endif
-        IPAddress broadcast_ip = IPAddress(local_ip[0], local_ip[1], local_ip[2], 0xFF);
+        broadcast_ip = IPAddress(local_ip[0], local_ip[1], local_ip[2], 0xFF);
       }
       for (int i=0; i < numLogValues; i++) {
         int outBufLen = 0;
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -68,6 +68,7 @@
  *        - ATTENTION: New categories for LMU64 and RVD/RVP controllers due to their different numbering schemes. Will be filled over time. PPS and sensor categories have moved up by two.
  *        - ESP32: OTA now uses system-wide HTTP AUTH authentication credentials
  *        - Improved built-in chart display (/DG), new configuration definement #define USE_ADVANCED_PLOT_LOG_FILE - thanks to Christian Ramharter
+ *        - Optional logging via UDP broadcast added (configurable, same parameters and format as in SD card logging)
  *        - Lots of bugfixes
  *       version 2.0
  *        - ATTENTION: LOTS of new functionalities, some of which break compatibility with previous versions, so be careful and read all the docs if you make the upgrade!
@@ -511,7 +512,7 @@ UserDefinedEEP<> EEPROM; // default Adresse 0x50 (80)
   #include <esp_task_wdt.h>
   #include <EEPROM.h>
   #include <WiFiUdp.h>
-WiFiUDP udp;
+WiFiUDP udp, udp_log;
   #if defined(ENABLE_ESP32_OTA)
     #include <WebServer.h>
     #include <Update.h>
@@ -527,10 +528,10 @@ EEPROMClass EEPROM_ESP((const char *)PSTR("nvs"));
 #else
   #ifdef WIFI
     #include "src/WiFiSpi/src/WiFiSpiUdp.h"
-WiFiSpiUdp udp;
+WiFiSpiUdp udp, udp_log;
   #else
     #include <EthernetUdp.h>
-EthernetUDP udp;
+EthernetUDP udp, udp_log;
   #endif
 #endif
 
@@ -5685,7 +5686,7 @@ void loop() {
 #endif
 // logged parameters
           #ifdef LOGGER
-            printFmtToWebClient(PSTR(",\r\n  \"logvalues\": %d,\r\n  \"loginterval\": %d,\r\n  \"logged\": [\r\n"), LoggingMode & CF_LOGMODE_SD_CARD, log_interval);
+            printFmtToWebClient(PSTR(",\r\n  \"loggingmode\": %d,\r\n  \"loginterval\": %d,\r\n  \"logged\": [\r\n"), LoggingMode, log_interval);
             not_first = false;
             for (i=0; i<numLogValues; i++) {
               if (log_parameters[i] > 0)  {
@@ -6208,6 +6209,7 @@ void loop() {
                 printyesno(logTelegram & LOGTELEGRAM_UNKNOWN_ONLY);
               }
               break;
+            // note: no use case recognized for temporary UDP logging on/off
             case '=': // logging configuration: L=<interval>,<parameter 1>,<parameter 2>,...,<parameter20>
             {
               char* log_token = strtok(p,"=,");  // drop everything before "="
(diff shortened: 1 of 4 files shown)


DE-cr:
Github says "All checks have failed".
Hmm, the code compiles perfectly in my Arduino 1.8.19 IDE.

Could it be that github requires preprocessor defines to start in the first column?
I placed them further to the right, for improved readability. I guess I'll move them to the very left and try again...

DE-cr:
Problems with github checks fixed:
- there was an erroneous space in a variable's type name (for non-ESP32 systems, which is why I didn't see it in my testing)
- the compiler used in my Arduino IDE seems to be "more forgiving" (i.e. wrong?) regarding variable scope
- (preprocessor directives with leading spaces are fine, as expected)

BSB_LAN/BSB_LAN.ino
```diff
@@ -6564,49 +6566,62 @@ void loop() {
...
-
-      if (dataFile) {
-        for (int i=0; i < numLogValues; i++) {
-          int outBufLen = 0;
-          if (log_parameters[i] > 0) {
-            outBufLen += sprintf_P(outBuf + outBufLen, PSTR("%lu;%s;%d;"), millis(), GetDateTime(outBuf + outBufLen + 80), log_parameters[i]);
-#ifdef AVERAGES
+      File dataFile;
+      if (LoggingMode & CF_LOGMODE_SD_CARD) {
+        #if defined(ESP32)
```
fredlcore, line 6574: We have decided a while ago that Indention of preprocessor directives should be independent of code indention (and yes, this has not been changed in the whole codebase). Please adjust the indention for the preprocessor directives in the parts of the code that you submitted.

BSB_LAN/BSB_LAN.ino
```diff
@@ -6564,49 +6566,62 @@ void loop() {
...
+        }
+      }
+      IPAddress broadcast_ip;
+      if (LoggingMode & CF_LOGMODE_UDP) {
+        #ifdef WIFI
+          IPAddress local_ip = WiFi.localIP();
+        #else
+          IPAddress local_ip = Ethernet.localIP();
+        #endif
+        broadcast_ip = IPAddress(local_ip[0], local_ip[1], local_ip[2], 0xFF);
```
fredlcore, line 6595: (this is an example why indention of preprocessor directives has to be separated. Because now it looks like as if broadcast_ip... and IPAddress local_ip... are not in the same code level)

BSB_LAN/BSB_LAN.ino
```diff
@@ -6564,49 +6566,62 @@ void loop() {
...
-  uint64_t freespace = 0;
-  freespace = SD.totalBytes() - SD.usedBytes();
-#else
-  uint32_t freespace = 0;
-  freespace = SD.vol()->freeClusterCount();
-#endif
-  if ((LoggingMode & CF_LOGMODE_SD_CARD) && freespace > MINIMUM_FREE_SPACE_ON_SD) {
+  if ((LoggingMode & (CF_LOGMODE_SD_CARD|CF_LOGMODE_UDP))) {
     if (((millis() - lastLogTime >= (log_interval * 1000)) && log_interval > 0) || log_now > 0) {
-//    SetDateTime(); // receive inital date/time from heating system
```
fredlcore, line 6576: Please don't remove commented code lines just because you don't think they're necessary. They may or may not be reminders for me to follow up on other functions.

BSB_LAN/BSB_LAN.ino
```diff
@@ -6564,49 +6566,62 @@ void loop() {
...
+        broadcast_ip = IPAddress(local_ip[0], local_ip[1], local_ip[2], 0xFF);
+      }
+      for (int i=0; i < numLogValues; i++) {
+        int outBufLen = 0;
+        if (log_parameters[i] > 0) {
+          if (LoggingMode & CF_LOGMODE_UDP) udp_log.beginPacket(broadcast_ip, UDP_LOG_PORT);
+          outBufLen += sprintf_P(outBuf + outBufLen, PSTR("%lu;%s;%d;"), millis(), GetDateTime(outBuf + outBufLen + 80), log_parameters[i]);
+          #ifdef AVERAGES
             if ((log_parameters[i] >= BSP_AVERAGES && log_parameters[i] < BSP_AVERAGES + numAverages)) {
-             //averages
```
fredlcore, line 6587: Same as above. Don't remove comments just because you think they're not necessary.

BSB_LAN/localization/LANG_DE.h
```diff
@@ -191,6 +191,7 @@
 #define MENU_LINK_URL "https://1coderookie.github.io/BSB-LPB-LAN/kap05.html#51-url-befehle"
 #define MENU_TEXT_LOT "Telegramme loggen"
 #define MENU_TEXT_FSP "Freier Speicher auf SD Karte"
+#define MENU_TEXT_WSS "WIFI Signalstärke"
```
fredlcore, line 194: This is not used in our code and might confuse translators. Please remove from the PR.

fredlcore:
Thanks for the PR, please find my few comments in the code.

BSB_LAN/localization/LANG_DE.h
DE-cr: Sorry, that was a leftover from my personal version's implementation of https://github.com/fredlcore/BSB-LAN/issues/455

BSB_LAN/BSB_LAN.ino
DE-cr: Sorry, I wasn't aware of a coding style guideline for BSB_LAN, so I formatted the code section I worked on to make it readable to me.

Btw, consistent formatting could easily achieved by using a pretty printer on the source files. Back in the previous millenium, when I was more active in programming, I used indent, which seems to be still around (https://man.openbsd.org/indent.1). Personally, I would do that for "production" code, as well as using a code checker that is stricter than most compilers (e.g. https://docs.oracle.com/cd/E19205-01/820-4180/man1/lint.1.html), and follow its suggestions.

BSB_LAN/BSB_LAN.ino
DE-cr: Point taken. :)

BSB_LAN/BSB_LAN.ino
DE-cr: Understood, sorry.

BSB_LAN/BSB_LAN.ino
DE-cr: With two "AVERAGES" and one "Averages" in the code line directly above, I considered this "//averages" obvious, sorry.

DE-cr:
You may want to check out the following:
1. Does my edit in BSB_LAN.ino line 5689 follow that code section's intent?
2. Are you offended by my using "if (condition) short_code_line;" instead of "if (condition) {\n  short_code_line;\n}"?
3. Is the setPPS() function (which I did not touch) ok for you?:
3.1. It seems to be unused.
3.2. Its header comment says "Global resources used: pps_values[]", but its code also uses (reads) the following global variables: LoggingMode, numLogValues, log_parameters[].

BSB_LAN/BSB_LAN.ino
DE-cr: P.S. However, using a pretty printer on all source files would most likely destroy the "clever" indentation in my /DG implementation in html_strings.h

fredlcore: Yes, and that's why we have dismissed using pretty printers after similar discussions on other issues before.

fredlcore:
1. No it doesn't, but previous sections of the code don't either, so it's a bit hard to do indention correctly.
2. Personally, I prefer the second way even for single lines of code because with the first way, you don't see at first glance if there is an indention mistake. But various contributors have used this way before, so I don't mind.
3. It is used in pps_handling.h, but you are right, the function comments are updated poorly.

DE-cr:
Re 1.: I meant intent, not indentation: Does my code change do what the code should do? The original version of that line seemed to be wrong to me.
Re 3.: A common problem. Personally, I prefer no comments to misleading ones. ;)

BSB_LAN/BSB_LAN.ino
DE-cr: P.P.S. But at least with the venerable "indent" pretty printer, preserving "clever" formatting is easily done: Just embed those code sections in "/* *INDENT-OFF* */ ... /* *INDENT-ON* */". (The last time I used indent myself was in the previous millenium, so I had to look that one up.)

Personally, I would find a common coding style for projects with multiple contributors helpful, especially since it could be easily/automatically enforced. Btw, of all the coding style descriptions I've seen so far, the one I like best is https://www.kernel.org/doc/html/v4.10/process/coding-style.html - although I've clearly not been following it here.
As it is, BSB_LAN code would not get highest marks from me for coding style / maintainability - but neither does my own work. ;)

fredlcore:
Re 1: Right, I was wondering about that, too. Could you have a look at that, please, @dukess?

BSB_LAN/BSB_LAN.ino
fredlcore: Well, I don't care for other people's marks on coding style, so I guess that's that ;)...

DE-cr:
This may be worth mentioning: In my testing of the UDP broadcasts, every now and then some packets got lost in transit (server=BSB_LAN to client=Perl@Linux). As far as I'm concerned, that's ok/expected: This is UDP broadcasting, not TCP data exchange with handshake.

However, I did NOT notice lost packets with the custom code for UDP broadcasting that I've been using for my https://github.com/DE-cr/BSBmonCR

The difference in the implementation is:
- six packages (=parameters) sent in a row in this implementation here, with each package about 60-80 bytes vs.
- one package/parameter (size = 9 Bytes) every two seconds in the custom BSB_LAN code for BSBmonCR

(Six parameters to be logged/sent was my test set.)

fredlcore:
Thanks, also for the observation regarding dropped packages. Yes, we can't expect UDP to be reliable, but it would be great to figure out if this is a general behaviour now or whether it does not exist with fewer parameter (just 1 or 2). Also, what is "every now and then"? Every five minutes, a few times per hour or once a day?
Other than that: Have you finished everything you wanted to do? If so, I would merge the PR. And thanks a lot for your efforts!

DE-cr:
Yes, I've finished what I wanted to do.

As for the package loss rate: I've now run a test with my standard six parameters (8830,8770,8700,8005,8003,8001) and a log intervall of 5 seconds. After 1000 log events, the result is as follows (number of parameters/packages received for each log event, with 6 (=full set) replaced by a period for better readability):

`...........................................5......
..................................................
..................................................
..................................................
..................................................
..................................................
.................5................................
.................................1................
..................................................
..................................................
..................................................
..................................................
..................................................
.....5...................5..............5.........
..................................................
..................................................
..................................................
..................................................
.........................................5........
..................................................`

This means that in this test, `6*1+1*5==11` out of `1000*6` packages got lost, i.e. less than 0,2 percent.

FYI: My BSB-LAN unit's wifi signal is reliable, but definitely not the strongest, as the unit is placed within the boiler's housing and on a different floor than the router. Also, its signal strength is not 100% stable. This may or may not explain the package loss rate here.

DE-cr:
...and in another test run (same settings):
`..................................................
..................................................
..................................................
..................................................
..................................................
..............................5...................
..................................................
..................................................
...................................3..............
..................................................
..................................................
..................................................
..................................................
..................................................
...........................5......................
..................................................
..................................................
..................................................
..................................................
..................................................`
I.e. `5/6000` packages lost <= 0,1 %

Here's my test script, in case you want to run your own test on your system(s):
```
#!/usr/bin/perl -w \

use strict;
use warnings;
use IO::Socket::INET;

++$|;
my $socket = IO::Socket::INET->new( LocalPort=>6502, Proto=>'udp' );
my $prev = time;
my $now;
my $n = 0;
my $nn = 0;

while ( $socket->recv( $_, 99 ) ) {
  $now = time;
  if ( $now - $prev > 2 ) {
    print $n if $nn;
    print "\n" unless !$nn || $nn % 50;
    exit if ++$nn > 1000;
    $n = 0;
    $prev = $now;
  }
  ++$n;
}
```

fredlcore:
Great, thanks a lot for this intensive testing! This looks good (at least acceptable) to me, so I'll merge this now :)...

PR #478, opened 2022-08-30
Some fixes

dukess:
Fix modf() for Arduino Studio
/Q output fix


BSB_LAN/include/get_cmdtbl_values.h
```diff
@@ -1,65 +1,25 @@
 
-uint32_t get_cmdtbl_cmd(int i) {
-  uint32_t c = 0;
-  int entries1 = sizeof(cmdtbl1)/sizeof(cmdtbl1[0]);
-  int entries2 = sizeof(cmdtbl1)/sizeof(cmdtbl1[0]) +  sizeof(cmdtbl2)/sizeof(cmdtbl2[0]);
-  if (i < entries1) {
+inline uint32_t get_cmdtbl_cmd(int i) {
```
fredlcore, line 2: Just out of curiosity: Why is `inline` necessary here?

fredlcore:
Thanks a lot!

BSB_LAN/include/get_cmdtbl_values.h
dukess: In the old books, there was a recommendation to do small functions "inline" since the overhead of calling a function was more than just embedding a piece of code. I tried both options today - there is no difference. It seems that the compiler itself decides when to embed the code and when to make a function call.

fredlcore: Thanks! Again learned something new (or old ;) )

fredlcore:
Compiling here goes well, but the CI job fails, could you have a look at the errors? Thanks!

fredlcore:
Sorry, wrong repo ;)...

PR #491, opened 2022-10-06
Fix: parameter numbers in logging w/o trailing ".0"

DE-cr:
...or was that intentional?


fredlcore:
Yes, the decimal part of the parameter (i.e. 8300**.0**) is intentional to be consistent with the webinterface.

Issue #513, opened 2022-11-12
[BUG] Home Assistant


proasnet:
**Before submitting a bug report, please make sure that you have checked [chapter 14](https://1coderookie.github.io/BSB-LPB-LAN_EN/chap14.html) of the manual** ("Problems and their Possible Causes"). Most problems that are reported to us can be solved by going through the solutions provided there.

**BSB-LAN Version**
3.0.1-20221101130752

**Architecture**
Arduino Due

**Bus system**
BSB

**Describe the bug**
Read MQTT data in Home Assistant, because integration from @liudger not working now, I tried add an easy sensor.
IP adress of broker, username and password is right.
Device ID and prefix is as default BSB-LAN

**To Reproduce**
1. Go to manual, chapter 8.11
2. Click on configuration.yaml and edit
3. paste
```
sensor:
  - platform: mqtt
    state_topic: "bsb-lan/8310"
    name: BSB-LAN flow temperature
    unit_of_measurement: °C
    device_class: temperature
```
This is not valid configuration.
I tried a new format
```
mqtt:
  sensor:
    - name: "BSB-LAN flow temperature"
      state_topic: "bsb-lan/8310"
      unit_of_measurement: "°C"
      device_class: temperature
```
5. See state of entity, is still unknown in both cfgs

**Logs**
LOG from MQTT Broker in HA:
`2022-11-12 15:52:18: New connection from 192.168.0.150:51415 on port 1883.`

State by BSB Lan
http://192.168.0.150/8310
8310.0 Diagnostics producer - Boiler temp actual value: 37.7 °C

**Additional context**
I am doing some bad, but I dont know what...

proasnet:
I tried this, and not working too.

```
rest:
  sensor:
    - name: "BSB-LAN status"
      resource: http://192.168.0.150/JQ=700,1600,8000,8003,6705,710,712,720,721,1610,1612,5400
      method: POST
      username:
      password:
      scan_interval: 60
      value_template: "{{ value_json['6705'].value }}"
      json_attributes:
        - "700"
        - "1600"
        - "6705"
        - "8000"
        - "8003"
        - "710"
        - "712"
        - "720"
        - "721"
        - "1610"
        - "1612"
        - "5400"
```

The check configuration in HA return this
`Invalid config for [rest]: [resource] is an invalid option for [rest]. Check: rest->rest->0->sensor->0->resource. (See /config/configuration.yaml, line 40). `

If I try use resource url in browser, I get this
```
[...]

PR #514, opened 2022-11-12
Fix for Issue #513. Correct timestamps for log files. Fix for new SdFat

dukess:
Fix for Issue https://github.com/fredlcore/BSB-LAN/issues/513
Adaptive (int/float) program numbers printing for web interface, MQTT, UDP/file logging.
Due: callback for SdFat library added for setup correct timestamps on log files.
ESP32: TimeLib library wrapper added. This wrapper can't be used with TimeLib in same time on ESP32. Wrapper will allow to set correct timestamps on log files.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -3858,7 +3858,7 @@ char *build_pvalstr(bool extended) {
     if(roundf(decodedTelegram.prognr * 10) != roundf(decodedTelegram.prognr) * 10)
       len+=sprintf_P(outBuf, PSTR("%4.1f "), decodedTelegram.prognr);
     else
-      len+=sprintf_P(outBuf, PSTR("%d "), roundf(decodedTelegram.prognr));
+      len+=sprintf_P(outBuf, PSTR("%d "), (int)roundf(decodedTelegram.prognr));
 
     len+=strlen(strcpy_PF(outBuf + len, decodedTelegram.catdescaddr));
     len+=strlen(strcpy_P(outBuf + len, PSTR(" - ")));
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -3004,7 +3004,13 @@ void generateJSONwithConfig() {
  *   none
  * *************************************************************** */
 char *GetDateTime(char *date) {
+#if defined(ESP32)
+  struct tm now;
+  getLocalTime(&now,0);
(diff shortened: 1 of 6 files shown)


mirkolenz:
Thank you for coming up with a solution that fast. From what I understood in the issue, you propose the following:

- Parameters ending in .0 will be published without the suffix, thus 700.0 would be available under bsblan/700
- Parameters ending in any other digit would be published with the digit, for instance bsblan/700.1

I hope I got that right. While I like the solution overall, I think it may be confusing to users: The web interface always shows the trailing zero, meaning that there is a mismatch if one tries to copy the parameter IDs to their config.

At the same time, I am not sure what the best solution would be here. Duplicate messages are not good and a new setting for this is overkill. I just wanted to raise awareness for this fact.

dukess:
Yes, your meaning correct.
I changed program numbers behavior for web interface.

fredlcore:
@dukess: Could you say also how you changed program numbers in the web interface? Just in case @1coderookie has screenshots in the manual that then are no longer correct.

mirkolenz:
@dukess Great, that seems like the most elegant solution 😄

dukess:
@fredlcore it would be look like 2.0 version:
изображение

изображение

fredlcore:
That looks great, thanks! Does this span across all modes (web interface, MQTT, JSON)?

dukess:
Now everything seems to be ready.

fredlcore:
Great, thanks a lot!

BSB_LAN/BSB_LAN.ino
```diff
@@ -3839,11 +3855,11 @@ char *build_pvalstr(bool extended) {
   int len = 0;
   outBuf[len] = 0;
   if (extended && decodedTelegram.error != 257) {
-#if !(defined ESP32)
-    len+=sprintf_P(outBuf, PSTR("%4.1f "), decodedTelegram.prognr);
-#else
-    len+=sprintf_P(outBuf, PSTR("%4.1f "), decodedTelegram.prognr);
-#endif
+    if(roundf(decodedTelegram.prognr * 10) != roundf(decodedTelegram.prognr) * 10)
+      len+=sprintf_P(outBuf, PSTR("%4.1f "), decodedTelegram.prognr);
```
fredlcore, line 3859: Not sure right now: Does %4 mean that program numbers are all 4 digits before the dot? Because we have everything from 1 to 5 digits?

dukess: "4" can be deleted because has no effect on the output

fredlcore: Thanks!

Issue #515, opened 2022-11-13
[BUG] JSON error from JI request


liudger:
**BSB-LAN Version**
"3.0.1-20221101130752"

**Architecture**
 Arduino Due

**Bus system**
BSB

**Describe the bug**
When doing a request for device info `URL/JI` and no SD card installed. The response has a malformed json.

**To Reproduce**
1. no SD card in bsblan device
2. do the JI query
3. See error
```{
    "name": "BSB-LAN",
    "version": "3.0.1-20221101130752",
    "hardware": "Due",
    "freeram": 81271,
    "uptime": 1664668,
    "MAC": "00:80:41:19:69:90",
    "freespace": ,
    "bus": "BSB",
    "buswritable": 0,
    "busaddr": 66,
    "busdest": 0,
    "monitor": 0,
    "verbose": 1
}
```
freespace has no value, should be zero?

**Logs**
If possible, attach or copy/paste a log of the Arduino IDE's serial monitor when performing the above actions. If the error is reproducible, do so after pressing the reset button on the board.

**Expected behavior**

**Screenshots**

**Desktop (if applicable, please complete the following information):**
 - OS: [e.g. MacOS 11.1, Windows 11]
 - Browser [e.g. Chrome, Safari]
 - Browser version [e.g. 22]

**Smartphone (if applicable, please complete the following information):**
 - Browser [e.g. Safari, FireFox]
 - Browser version [e.g. 22]

**Additional context**

dukess:
Question: LOGGER was enabled?

liudger:
Not sure. I will check. Still there should be valid json. 0 or Nan?

dukess:
In https://github.com/dukess/BSB-LAN/commit/bd43d5a68482ce44da86fedc332c94afd3e46c33 i propose to remove "freespace" field when SD card not initialized.

fredlcore:
Hm, I don't know much about JSON parsing, but would it create problems if a field is queried if it's not present? Otherwise I'd suggest `NaN` if the SD card is not present because `0` could also mean that the SD card is full.

dukess:
did that

fredlcore:
Thanks!

liudger:
I updated to the latest and the build failed. I had turned off some defines. And this caused a build issue. (fs_date was not defined) When I enabled most defines the build worked. Maybe we need to test this better? Or hide defines?

Also the issue was now solved!

fredlcore:
[...]

PR #517, opened 2022-11-14
Set of fixes

dukess:
Fix for https://github.com/fredlcore/BSB-LAN/issues/515
Adaptive (int/float) program numbers printing for /Q command
Fix for wrong month and weekday numbers

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -5671,15 +5671,15 @@ void loop() {
             printToWebClient(BSB_VERSION);
             printToWebClient(PSTR("\",\r\n  \"hardware\": \""));
             printDeviceArchToWebClient();
-            printFmtToWebClient(PSTR("\",\r\n  \"freeram\": %d,\r\n  \"uptime\": %lu,\r\n  \"MAC\": \"%02hX:%02hX:%02hX:%02hX:%02hX:%02hX\",\r\n  \"freespace\": "), freeRam(), millis(), mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+            printFmtToWebClient(PSTR("\",\r\n  \"freeram\": %d,\r\n  \"uptime\": %lu,\r\n  \"MAC\": \"%02hX:%02hX:%02hX:%02hX:%02hX:%02hX\""), freeRam(), millis(), mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
 #if defined LOGGER || defined WEBSERVER
 #if !defined(ESP32)
             uint32_t freespace = SD.vol()->freeClusterCount();
             freespace = (uint32_t)(freespace*SD.vol()->sectorsPerCluster()/2048);
-            printFmtToWebClient(PSTR("%d"), freespace);
+            printFmtToWebClient(PSTR(",\r\n  \"freespace\": %d"), freespace);
 #else
             uint64_t freespace = SD.totalBytes() - SD.usedBytes();
-            printFmtToWebClient(PSTR("%llu"), freespace);
+            printFmtToWebClient(PSTR(",\r\n  \"freespace\": %llu"), freespace);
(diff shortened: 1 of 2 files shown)


1coderookie:
FYI: Someone just complained about not being able to compile for ESP32, error message says something about "esp32_time.h" and so on. I don't know if that has something to do with the last PR merged two days ago, but just in case I wanted to let you know.. ;)
https://forum.fhem.de/index.php/topic,29762.msg1245538.html#msg1245538

dukess:
src/Time directory is superfluous for ESP32

fredlcore:
But it will only work on ESP32, not on Arduino, right? If so, users would have to delete the library they're not using, which is a bit of a hassle. What was again the benefit of switching from TimeLib to esp32_time?

fredlcore:
Compiling for Arduino works fine without having to delete esp32_time.h, but not when compiling for ESP32.

dukess:
Same story as WiFiSpi: this library can't build for ESP32.
Benefit is small but exist: esp32_time is wrapper for POSIX time functions which native used in ESP32. May be RTC will be more accuracy. Main benefit (for me): files that program records will have the correct date and time of the change.

1coderookie:
Ok, I don't really understand everything here - so please let me know at the end what I have to adjust/note/change within the manual ;) thx

dukess:
>Compiling for Arduino works fine without having to delete esp32_time.h, but not when compiling for ESP32.
Yep. I added directives 
#if defined(ESP32)
#endif
so it never be compiled on Arduino.

dukess:
> please let me know at the end what I have to adjust/note/change within the manual ;) thx

Sure!

fredlcore:
But thesel ines here should have the effect that TimeLib also never compiles on ESP32:
```
#if defined(ESP32)
  #include "src/esp32_time.h"
#else
  #include "src/Time/TimeLib.h"
#endif
```
The strange thing with the compilers is that they seem to compile every library in the directory structure, even if the library is nowhere included in the code. Don't know why this is done and how to prevent this, but for the time being, @1coderookie please add `TimeLib` to the folders that ESP32 users have to remove before compiling, similar to ArduinoMDNS and WiFiSPI. Thanks!

1coderookie:
> @1coderookie please add `TimeLib` to the folders that ESP32 users have to remove before compiling, similar to ArduinoMDNS and WiFiSPI. Thanks!

I'll do it asap, thx (..so I guess there'll be coming a few questions/reports about this in the near future, as users tend to not read the manual again after successfully flashing once ;) )

fredlcore:
Ah, stop the press ;) - now I understood what @dukess meant with adding the conditional defines, I was looking at the wrong spot. I have modified TimeLib and now it compiles fine here. Will do the same with ArduinoMDNS and WiFiSPI and then it will hopefully no longer be necessary to remove these directories when compiling for ESP32. We just have to remember to add these conditionals into the libraries' source code once we update them to a newer version.

dukess:
> The strange thing with the compilers is that they seem to compile every library in the directory structure, even if the library is nowhere included in the code. Don't know why this is done and how to prevent this, but for the time being, 

Sad but as is.

1coderookie:
Wait - WHAT? Not necessary anymore in the near future to delete those folders and I don't need to add TimeLib to WiFiSpi etc? Well THATS good news! :partying_face: 
Ok so let me know when to remove the note about WiFiSpi/ArduinoMDNS then..thx ;)

fredlcore:
Ok, I added conditionals to all the Arduino-specific libraries so they won't compile. @1coderookie: You can remove the instructions to remove ArduinoMDNS and WifFiSpi from the manual. It won't hurt if people still delete them (because they have an old manual on their PC), but it won't be necessary anymore with the fix I just added :)...

1coderookie:
Sweet! Will do it asap..

fredlcore:
Thanks!

Issue #520, opened 2022-11-17
[BUG] Exception in mqtt_sendtoBroker for mqtt_mode 3


fschaeck:
**BSB-LAN Version**
3.0.1-20221101130752

**Architecture**
ESP32 NodeMCU (Joy-It)

**Bus system**
BSB

**Describe the bug**
When using mqtt_mode 3 an exception is thrown AFTER publishing the json message - apparently due to a buffer overflow in outBuf. Using mqtt_mode 1 or 2 this exception is not thrown but the message is also significantly shorter.

Also in function mqtt_callback(...) in BSB_LAN/include/mqtt_handler.h the buffer overflow protection is reserving 4 bytes for the acknowledgement message, but 5 are needed (4 for the ACK_ prefix and one for the terminating nut character).

**To Reproduce**
Set mqtt_mode in BSB_LAN_config.h to 3, compile and upload sketch, watch the serial monitor to see the exception happening.

**Logs**
(These logs are taken from an already modified mqtt_handler.h to pin down the problem)
12:41:55.566 -> �f`ADY
12:41:55.598 -> Using settings from config file
12:41:55.598 -> Reading done.
12:41:55.630 -> EEPROM schema v.0 Program schema v.11
12:41:55.630 -> Address EEPROM option 0: 0
12:41:55.630 -> Address EEPROM option 1: 1
12:41:55.630 -> Address EEPROM option 2: 2
12:41:55.630 -> Address EEPROM option 3: 6
12:41:55.630 -> Address EEPROM option 4: 226
12:41:55.630 -> Address EEPROM option 5: 306
12:41:55.662 -> Address EEPROM option 6: 490
12:41:55.662 -> Address EEPROM option 7: 491
12:41:55.662 -> Address EEPROM option 8: 492
12:41:55.662 -> Address EEPROM option 9: 493
12:41:55.662 -> Address EEPROM option 10: 494
12:41:55.662 -> Address EEPROM option 11: 495
12:41:55.662 -> Address EEPROM option 12: 499
12:41:55.662 -> Address EEPROM option 13: 659
12:41:55.662 -> Address EEPROM option 14: 663
12:41:55.662 -> Address EEPROM option 15: 823
12:41:55.662 -> Address EEPROM option 16: 829
12:41:55.662 -> Address EEPROM option 17: 830
12:41:55.694 -> Address EEPROM option 18: 834
12:41:55.694 -> Address EEPROM option 19: 838
[...]

PR #521, opened 2022-11-17
Fixes in mqtt_handler.h

fschaeck:
An exception was thrown after publishing a json message in mqtt_mode 3. Apparently a buffer overlow in outBuf was responsible for the exception, since it did not happen for mqtt_mode 1 or 2, where the messages are significantly shorter. The refactored function mqtt_sendtoBroker() is not using outBuf anymore.

Also in the buffer overflow protection of function mqtt_callback(...) 4 bytes were reserved for creating the acknowledgement message from the original message, but 5 bytes are needed (4 for the ACK_ prefix and one for the terminating nul character).

This pull request is fixing issue #520

### DIFF
--- a/BSB_LAN/include/mqtt_handler.h
+++ b/BSB_LAN/include/mqtt_handler.h
@@ -58,10 +58,11 @@ void mqtt_sendtoBroker(float param) {
     // =============================================
     case 1:
       // use parameter code as sub-topic 
-      if(roundf(param * 10) != roundf(param) * 10)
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks a lot for reporting and fixing this bug :). Please use opening and closing curly brackets in if clauses etc. as we do in the rest of the code. Once formatting is in order, I'll merge this into master.

fredlcore:
Thanks a lot!

fschaeck:
Okay! I fixed the formatting for the whole file since I was at it.

PR #522, opened 2022-11-19
replacing "%.1f" with "%g". Fix  for /JC and ?JK commands

dukess:
"%.1f" was replaced with "%g". Code more readable, shorter (and faster?). Thanks to @DE-cr 
roundf() was replaced with (int)() where is it possible.
Fix for /JC and /JK commands (additional condition in printToWebClient_prognrdescaddr() may block parameter name printing) 
Tested on ESP32.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -995,15 +995,6 @@ void printHTTPheader(uint16_t code, int mimetype, bool addcharset, bool isGzip, 
   }
 }
 
-char *printProgNR(float prognr, char *buf){
-  if((((int)prognr) * 10) == ((int)(prognr * 10))) {
-    sprintf(buf, "%d", (int)prognr);
-  } else {
-    sprintf(buf, "%.1f", prognr);
-  }
-  return buf;
-}
-
 int recognize_mime(char *str) {
 //        if (strlen(dot)) {
   char mimebuf[32];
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks!

fredlcore:
We have a user reporting this behaviour where an unwanted parameter number appears after the parameter value:
unwanted param numbers
Is this related to this PR?

dukess:
Yes, I ran into this problem too. In addition, rounding does not work correctly.
Enrages.
I think it's better that I cancel the commit (except for a minor fix) and leave everything as it was before. Less beautiful, but it worked.

fredlcore:
Ok, could you do that because I don't know what minor fix should be kept? Thanks!

DE-cr:
The %g part looks ok to me, but there seems to be one printFmtToWebClient() to many.
(That is most of the changes _could_ be kept, but _something_ needs fixing.)

DE-cr:
I don't know the code intimately enough, but could it be that `desc` in `printToWebClient(cfg.desc);` (BSB_LAN.ino line 2732) has (somewhere before) been created to contain one param_no too many? Since `cfg` here is set via `memcpy(&cfg, &config[i], sizeof(cfg));` some lines before being used, that would mean `config[]` setting is the culprit. I have yet to figure out, where that happens.

DE-cr:
Sorry, no: cfg seems to be /C contents, not /Kx

dukess:
Tried to save most of changes. Did PR.

fredlcore:
Thanks!

PR #524, opened 2022-11-20
Fix for night commit

dukess:
fixed: rounding bug (roundf()->(int)() was reverted).
fixed: bug with additional program number printing


fredlcore:
Thanks!

PR #528, opened 2022-11-23
Fixed code for Web API /JK=<category-number>

fschaeck:
Using a category number outside the valid range would set cat_min and cat_max to arbitrary numbers and produce all kinds of crazy behaviour.

The category number is now checked before the array ENUM_CAT_NR is referenced and in case of an illegal category number 0 is used.


fredlcore:
Thanks!

PR #532, opened 2022-12-18
Optimization for commit 40fb48c8

dukess:


fredlcore:
This was just inserted by me for debugging purposes and is probably going away again, so I think we should keep it as it is for now, but thanks!

dukess:
Got it!

Issue #533, opened 2022-12-19
[BUG] bsb-lan fails to close http connection after delivering home page


DE-cr:
**BSB-LAN Version**
3.0.3-202212094171550 (github head version from 2022-12-19 07:51)

**Architecture**
ESP32 NodeMCU

**Bus system**
BSB

**Describe the bug**
On my system, the http connection is not closed properly after completely delivering the home page (http://bsb-lan/) contents (including the final `\</html\>\r\n\r\n`).
After a while, the client then complains (Firefox: "Fehler: Verbindung unterbrochen [...]") or just hangs (wget).
_This only seems to happen when "check for updates" is enabled!_

fredlcore:
Can you try and change line 110 in print_webpage.h from
flushToWebClient();
to
forcedflushToWebClient();
and see if the problem persists?

DE-cr:
Thank you for the quick reply - even with this less-than-substantial bug!
I've changed what you've suggested: The problem persists.

DE-cr:
Here's the WireShark (Linux) protocol: 
[2022-12-19_18-27_-cr_wireshark.zip](https://github.com/fredlcore/BSB-LAN/files/10261496/2022-12-19_18-27_-cr_wireshark.zip)

dukess:
BSB-LAN flush all bytes to client browser but doesn't connection closing after ~1000 msec.
May it rebooted in this time for some reason?

DE-cr:
Yes, that's it! Uptime in /C shows that it does reboot after delivering / contents. But why?

fredlcore:
What does SerMo log say? Is there a crash?

DE-cr:
There's no SerMo: The unit sits inside the boiler casing.
Should I try telnet monitoring?

fredlcore:
No point. In case there is a crash, how should the crash message be sent to the telnet client?

DE-cr:
@fredlcore, could you please remove the bogus `9` from http://bsb-lan.de/bsb-version.h and let me see if the problem persists?

fredlcore:
Why should this have anything to do with it and why don't you remove it yourself locally and test it then?

DE-cr:
The problem could be writing beyond the end of `version_number[15]` in `include/print_webpage.h: webPrintSite()`.
The version check in calling bsb-lan's root page reads http://bsb-lan.de/bsb-version.h, not my local copy.

DE-cr:
[...]

PR #543, opened 2023-03-04
Buffer overrun fix in version check on home page

DE-cr:
Fixes https://github.com/fredlcore/BSB-LAN/issues/533, at least on my system.


fredlcore:
Thanks!

PR #545, opened 2023-03-04
788 bytes code size reduction for svg_favicon

DE-cr:
1. ran the original svg_favicon code through https://www.svgminify.com/
2. manual re-<use> of svg code for the letter "B" (instead of code duplication)
3. removed line breaks in svg code


fredlcore:
Thanks!

PR #551, opened 2023-03-05
DEVELOPER_DEBUG - Very verbose debug output

dukess:
This option will dramatically slow down MCU performance. Useful for developers not for users.


fredlcore:
Thanks!

PR #542, opened 2023-03-04
/DG improvements for large datasets (introducing some new /D... commands)

DE-cr:
addresses https://github.com/fredlcore/BSB-LAN/issues/539

PR #554, opened 2023-03-10
Use 4 KB buffer in transmitFile() on ESP32

DE-cr:
... if available from the heap, otherwise falling back to the existing static (<=) 1 KB buffer
reason: performance improvement (>20% on an esp32 w/o SD card, using wifi)

An equivalent algorithm is already part of my https://github.com/fredlcore/BSB-LAN/pull/542, but only for /Dyyyy-mm-dd,YYYY-MM-DD, as /D and /Dn in that PR use transmitFile() directly (and therefore would also benefit from this PR here!).

The change has been done based on measuring /D times [s] on a datalog.txt of about 970 KB on my system (see above), comparing implementations using the original 1 KB buffer size and the new 4 KB buffer size for the transfer (file reading and data transmission over http):
```
      1 KB   4 KB   gain       
----+------+------+-----
#1    7.09   5.12   28 %
#2    6.92   5.18   25 %
#3    6.86   5.09   26 %
#4    6.80   5.51   19 %
#5    6.79   5.17   24 %
#6    6.80   5.02   26 %
#7    6.79   5.09   25 %
#8    6.76   5.44   20 %
#9    6.79   5.17   24 %
#10   6.80   5.14   24 %
----+------+------+-----
min   6.76   5.02   26 %
max   7.09   5.51   22 %
avg   6.84   5.19   24 %
```
I've created a new PR here, and have not included this in https://github.com/fredlcore/BSB-LAN/pull/542, to allow inclusion of this PR here, even if the other one is not accepted.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -4551,7 +4551,10 @@ uint16_t setPPS(uint8_t pps_index, int16_t value) {
  * *************************************************************** */
 void transmitFile(File dataFile) {
   int logbuflen = (OUTBUF_USEFUL_LEN + OUTBUF_LEN > 1024)?1024:(OUTBUF_USEFUL_LEN + OUTBUF_LEN);
-  byte *buf = (byte*)malloc(4<<10);  // try to use 4 KB buffer, for improved transfer rates
+  byte *buf = 0;
+#ifdef ESP32 // Arduino seems to have problems with the bigger, malloc'ed buffer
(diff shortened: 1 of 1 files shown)


DE-cr:
I'd appreciate others testing this change on systems I don't have available, most notably Arduino instead of ESP32, wired network connection instead of wifi, and using an SD card.

DE-cr:
Not doing this on Arduino now, because of https://github.com/fredlcore/BSB-LAN/pull/542#issuecomment-1474269236

PR #563, opened 2023-03-24
/DG plot readability improvements

DE-cr:
- tooltip values more visible
- cursor (and axes) slightly fainter
- defocused plot lines more visible

before:
grafik
grafik

after:
grafik
grafik


fredlcore:
Thanks, looks good, one wish that I would have (but not sure if that's easily doable) is that clicking on a parameter in the legend switches between hiding this parameter, showing it again and then showing only this parameter. I had several occasions in the last few days where I wanted to look at just one graph only and had to manually disable all the others and then re-enable them again (or rather reload the page). Yes, when I hover over a parameter, it becomes highlighted, but I can't move the mouse in order to see the actual values.
But it's just an idea, if it's more complicated doing it, then no worries...

DE-cr:
I'll look into it, but I expect this to take some digging deep into the c3.js sources, as I haven't been able to find an example for that, and it's not covered by the documentation, either. Therefore, I'd prefer to see this PR merged as is for now.

Btw, since the c3 library hasn't been maintained for a couple of years, I had a look billboard.js, which evolved from a c3 fork and is still active. However, as I have seen no noticeable gains for /DG, but slower execution times with billboard (and larger libraries to load), I've refrained from doing a PR for moving from c3 to billboard. This just fyi.

DE-cr:
> I had several occasions in the last few days where I wanted to look at just one graph only and had to manually disable all the others

After looking into the c3 sources, I'm happy to report that this use case is already supported by /DG as it is:
Alt-click on a legend entry enables that line only and disables all others.
One more bit of information for the manual, @1coderookie.
:)

PR #567, opened 2023-03-29
Support using local copies of *.js libs for /DG

DE-cr:


BSB_LAN/BSB_LAN_config.h.default
```diff
@@ -207,9 +207,19 @@ float ipwe_parameters[80] = {
...
 // - added zoom (mousewheel/pinch on plot) and pan capability (drag zoomed-in plot)
-// - saves over 4 KB in compiled BSB-LAN sketch
+// - selective plotting of data from big logs
 #define USE_ADVANCED_PLOT_LOG_FILE
 #define DEFAULT_DAYS_TO_PLOT "3" // must be a positive integer value in double quotes!
+// Log file plotting uses two JavaScript libraries, which are usually loaded from the web.
+// Should you prefer to use local copies instead, put the files from the data subdirectory onto your
+// bsb-lan unit's SD card (*), and provide their path names here. For browsers that support gzip
+// compression (what browser doesn't?), it is sufficient to copy the *.gz file versions to your
+// bsb-lan unit, but omit the ".gz" from the path names in the following lines!
```
fredlcore, line 217: This is not really clear: There is no ".gz" in the path names in the following lines.

DE-cr: Yes, because I've followed my own directions and put just x.js in those lines, although there's no x.js on my esp32, only x.js.gz. How would you word this?

The web server implementation I've found in bsb-lan.ino does not contain a routine to gzip files asked for by a client web browser on the fly. Instead, if asked for file x, it first looks for a file x.gz and delivers that, if the browser accepts gzip compression. However, that process is transparent to the client, which _must_ ask for file x.

DE-cr: Is this better?: https://github.com/fredlcore/BSB-LAN/commit/d9875d4ebfda92f4d2b0bd49e3ac03cb20fdf1be

fredlcore: Yes, now I got it ;)...

DE-cr:
Btw, using bsb-lan's built-in web server, I've noticed that LittleFS seems to be case sensitive with regard to file names. This may come as a surprise to users of Windows / (standard) MacOS file systems.

fredlcore:
Since everything in the config file is case sensitive, it shouldn't come as that much of a surprise (plus MacOS at least on terminal/shell level is case sensitive, too)

PR #577, opened 2023-04-22
Update brute-force.pl

nrobadey:
Remove the suffix in the URL


fredlcore:
The suffix is actually the passcode, so I'll leave it there so that people who use the passcode (most of them do) know that it has to be added here as well.

PR #584, opened 2023-05-20
Changing type of *_parameters arrays from 'float + float' to 'parameter'

dukess:
Some bugfixes for destination addresses output
Fix for parameter size calculation
parameter.dest_addr is 'short' now instead 'int'. Some bytes will be saved in EEPROM


fredlcore:
Is this ready to merge?

dukess:
Yes, ready

fredlcore ***@***.***> 31 мая 2023 г. 21:12:36 написал:

PR #594, opened 2023-10-14
Make uptime in /C more legible

DE-cr:
I always need a calculator to convert long ms values to `d+h:m:s`
With this PR in place, I don't:
grafik

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -2148,13 +2148,15 @@ void generateConfigPage(void) {
   printDeviceArchToWebClient();
   printToWebClient(PSTR("<BR>\r\n" MENU_TEXT_VER ": "));
   printToWebClient(BSB_VERSION);
-  printToWebClient(PSTR("<BR>\r\n" MENU_TEXT_RAM ": "));
-#ifdef WEBCONFIG
-  printFmtToWebClient(PSTR("%d " MENU_TEXT_BYT "<BR>\r\n" MENU_TEXT_UPT ": %lu<BR>\r\n"), freeRam(), millis());
-
-#else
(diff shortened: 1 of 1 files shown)


DE-cr:
P.s. I forgot to mention: I do consider this ready to be merged into BSB-LAN's master branch. :)

fredlcore:
Good idea, but please add it in a way that this information is printed in a new line (but without <BR> so that visually, it remains in the same line). Background: People who parse the current Uptime line using regex might catch everything after the colon and therefore wouldn't get a number back. If it's in a new line, this problem would not occur.

DE-cr:
That would be an overly simple parser. ;)

What do you mean by "(but without
so that visually, it remains in the same line)"?

fredlcore:
Well, that's how I parse the HTML output of BSB-LAN. If it is clear how the rest of the line looks like, why should I create an overly complex regex? That would be a waste of time and resources.
What I meant is: Add a crlf at the beginning of the line, but don't add a <BR>.

DE-cr:
Done, thanks for suggesting this:
```
</form>
<BR>Hardware: ESP32<BR>
Version: 3.2.2-20230606224119<BR>
Freier Speicher: 215548 Bytes<BR>
Uptime: 78504
ms = 0d+00:01:18.504<BR>
1-Wire Bus Pins: 
0, Sensoren: 0
<BR>
<BR><BR>
Installierte Module: <BR>
```
This didn't occur to me, as I usually try to layout my html similar to its expected rendering.

As for the regex: I personally wouldn't consider `\d+` "overly complex", but safer than `.+` (or even `.*`).
...and it could also be used to parse `Freier Speicher: 215548 Bytes<BR>`, e.g.
;)

fredlcore:
Good for you.

PR #598, opened 2023-11-10
Stop serving client that is already gone

DE-cr:
...at least for potentially long lasting actions, e.g. /D or /K24.
This avoids unnecessarily blocking bsb-lan for other tasks (including new client requests).

Succesfully tested on ESP32 with wifi connection, using the url commands given above.
I consider this ready for merging.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -5704,7 +5704,7 @@ void loop() {
           } else {
             printToWebClient(PSTR(MENU_TEXT_QFA "!"));
           }
-          for (int x=0;x<10;x++) {
+          for (int x=0; x<10 && client.connected(); x++) {
             if (found_ids[x]==0xFF) {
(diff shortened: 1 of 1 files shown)


DE-cr:
P.s. now also addressing /QD handling. (Tested on my system.)

fredlcore:
Great idea, thanks! Does this also mean that you can abort let's say a long query of parameters (such as /0-9999) by hitting the "Stop" button on the browser and start a new query right after? Because that has often confused users who were wondering why BSB-LAN was still not responding for minutes until the query in the background was completed.

DE-cr:
Yes, tested right now. :)

DE-cr:
Btw, I've done my testing by calling the url commands that usually take a long time, then immediately going back to the previous page in the browser (alt + left), which cancels the client connection and creates a new one.

fredlcore:
Great, that's really helpful!

PR #606, opened 2024-01-02
New stl files for Olimex POE-ISO Board

V4n1X:
Version 3 (First public release)


fredlcore:
Thanks, can you explain briefly what you have changed and why?

V4n1X:
Changed the box size in length for the Olimex POE-ISO because it's 18mm longer then the not ISO variant.
Also changed the corresponding holes to fit perfectly for the variant incl. the holes for the BSB-LAN attachment board.
(Thats why i marked it with version 3, needed 2 prints because it didnt fit on the first two perfectly).

Everything is based on the original not ISO variant STL file.

Can upload also some real world pictures for verificationy, if needed.

fredlcore:
Thanks, just wanted to make sure that it is what I expected it to be :)...

fredlcore:
Thanks!

PR #620, opened 2024-02-25
Fix: overwrite buffer in cleanupDatalog()

dukess:


dukess:
BTW: should we define separate buffers for client, debug and file outputs?

fredlcore:
Thanks! I have no Due at the moment where I could check whether this (seriously) affects free RAM, but if the buffers are just created temorarily, then it shouldn't be an issue, right?

DE-cr:
The use of a temporary char* here doesn't make a difference.
Whether flushToWebClient() does here, I cannot easily tell, but I would expect that right before processing a new url command, that buffer should be flushed.
Anyway, the code change should not hurt, except for a few bytes more in the size of bsb-lan's binary. ;)

PR #625, opened 2024-03-11
[IMP] github workflow: test options

jbaudoux:
Detect compilation errors based on available options

### DIFF
--- a/.github/workflows/platformio-esp32-options.yaml
+++ b/.github/workflows/platformio-esp32-options.yaml
@@ -1,0 +1,140 @@
+name: PlatformIO CI ESP32
+
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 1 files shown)


fredlcore:
I'm most likely not going to accept this because it will significantly increase completion of the workflow. One run usually takes around 40 seconds to one minute. This change would result in an approx. 20 minute run - per architecture.

fredlcore:
You can run time tests in a fork of yours and post the difference here, then I can decide.

fredlcore:
Some are, but to disable them is kind of a relict from the Arduino Mega era where at the end we needed to save memory by disabling some functions. This is no longer necessary with the more modern platforms where configuration is done mostly through the webinterface.

PR #627, opened 2024-03-13
Re-enable use of Arduino 1.8.19 for ESP32

DE-cr:
With https://github.com/fredlcore/BSB-LAN/commit/0d59fc1f9ecd51dc611259d1fb66d4427615d1c3 compiling for ESP32 using Arduino IDE 1.8.19 is not possible anymore because the types `uint64_t` and `String` need to be defined.


fredlcore:
Thanks, but it's not only relevant for ESP32, but for all, so I moved the existing include above the forward declarations here. Thanks anyways.

PR #637, opened 2024-03-24
Wait USB Serial connection to establish before sending debug

jbaudoux:
On my device, on power on, the esp starts faster than the usb-ttl chip.
It's start outputting useful init debug info while the serial connection is not ready yet.
It needs between 500ms and 1sec to initialize. I added a delay in the setup before sending the first message. As this is only at power on, I don't think such delay has any drawback.


fredlcore:
It does not apply to any of our supported boards. You can add the required commands to BSB_LAN_global.h or BSB_LAN_custom_setup.h instead.

PR #641, opened 2024-03-31
[WIP] Add MQTT logging

jbaudoux:
Forward messages to MQTT broker

Relates to https://github.com/fredlcore/BSB-LAN/issues/442

In your `BSB_LAN_config.h`, set 
* byte LoggingMode = 4;  // CF_LOGMODE_MQTT
* int logTelegram = LOGTELEGRAM_ON

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -2886,14 +2886,17 @@ void generateJSONwithConfig() {
  *  Pass parameters:
  *   byte *msg pointer to the telegram buffer
  * Parameters passed back:
- *   none
+ *   char * pvalstr
  * Function value returned:
  *   none
  * Global resources used:
  *   log_parameters
  * *************************************************************** */
-void LogTelegram(byte* msg) {
-  if (!(logTelegram & LOGTELEGRAM_ON) || !(LoggingMode & CF_LOGMODE_SD_CARD)) return;
+void LogTelegram(byte* msg, float query_line) {
+  printTelegram(msg, query_line);
+  // printTelegram has populated decodedTelegram
+  if (logTelegram & LOGTELEGRAM_OFF) return;
+  if (!(LoggingMode & (CF_LOGMODE_SD_CARD | CF_LOGMODE_MQTT))) return;
   File dataFile;
   uint32_t cmd;
   int i=0;        // begin with line 0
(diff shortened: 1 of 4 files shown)


fredlcore:
Is there really no way that you can work on your contributions in your fork and only submit a PR when it's ready for review? Then these "draft PRs" wouldn't clutter my project's PR list which is basically my main ToDo list for this project.

And again: I don't entertain any questions before there is an explanation on what is planned to do and how it is planned to do it. Other than that, the questions you ask show that you haven't even made it to chapter 5 in the manual.

fredlcore:
No one - including me - knows what parameters you are referring to when you talk abount parameter numbers > 10020...

fredlcore:
Even though it's still in draft, please take note of the following:
- Have a look at the function `parsingStringToParameter`.
- We specifically introduced the `parameter` struct for easier handling of parameters as well as destinations. Do not introduce your own kind of handling of that by removing code that explicitly uses this struct, unless you have a good reason to do so.

Again, please explain what you plan to do before wasting your own time on a PR that I may not accept in the end.
I appreciate if someone takes the time to refactor this part of the code which has grown organically too entangled, but if we do such a major move, it should be done solidly:
- `query()` should be removed from `mqtt_sendtoBroker()`. If the former is executed before the latter, we will have `decodedTelegram` populated with all relevant information, so it can be published right away.
- `query()` should be called not only with the line number but a `parameter` struct. Having also the destination device ID in the function would enable us to move a lot of the temporary switching of destination devices to the `query()` function, as well as (optionally) call `mqtt_sendtoBroker()` which requires a `parameter` struct to be passed to it.
- `printTelegram` also uses `decodedTelegram`, so `mqtt_sendtoBroker()` could easily be called from there as well. There is also quite a significant amount of overlap between `LogTelegram()` and `printTelegram()`, and it might be worthwhile to bring together all log functionalities (debug output, SD card, MQTT, UDP) together in one function (or one with several sub functions).

BSB_LAN/BSB_LAN.ino
```diff
@@ -2956,8 +2964,13 @@ void LogTelegram(byte* msg) {
     default: logThis = false; break;
   }
   if (logThis) {
-    dataFile = SDCard.open(journalFileName, FILE_APPEND);
-    if (dataFile) {
+    if (LoggingMode & CF_LOGMODE_MQTT) {
+      printlnToDebug(PSTR("Send to MQTT"));
+      if ((decodedTelegram.prognr != -1) and (!is_query)) {
+        mqtt_sendtoBroker(-1);
```
fredlcore, line 2970: How do you know that the destination address is always the default destination address (that's what -1 stands for in `param.dest_addr`)?

BSB_LAN/BSB_LAN.ino
```diff
@@ -2906,13 +2909,18 @@ void LogTelegram(byte* msg) {
...
 #else
-  if (totalBytes()-usedBytes() < minimum_SD_size) return;
+    if (totalBytes()-usedBytes() < minimum_SD_size) return;
 #endif
+    dataFile = SDCard.open(journalFileName, FILE_APPEND);
+  }
 
+  bool is_query = false;
   if (bus->getBusType() != BUS_PPS) {
+    if (msg[4+(bus->getBusType()*4)]==TYPE_QUR) is_query = true;
```
fredlcore, line 2923: What about TYPE_QINF, TYPE_ACK, TYPE_NACK etc., do you want to log them as well? Do you know the state of `decodedTelegram` in these cases?

BSB_LAN/BSB_LAN.ino
```diff
@@ -2906,13 +2909,18 @@ void LogTelegram(byte* msg) {
   int data_len;
   float dval;
   float line = 0;
+  if (LoggingMode & CF_LOGMODE_SD_CARD) {
 #if !defined(ESP32)
-  if (SDCard.vol()->freeClusterCount() < minimum_SD_size) return;
+    if (SDCard.vol()->freeClusterCount() < minimum_SD_size) return;
```
fredlcore, line 2914: Wrong indention

BSB_LAN/BSB_LAN.ino
```diff
@@ -2906,13 +2909,18 @@ void LogTelegram(byte* msg) {
...
 #else
-  if (totalBytes()-usedBytes() < minimum_SD_size) return;
+    if (totalBytes()-usedBytes() < minimum_SD_size) return;
```
fredlcore, line 2916: Wrong indention

fredlcore:
This has become obsolete now, I think. Thanks for the ideas anyway.

PR #645, opened 2024-04-07
Add BSB-LAN watermark to /DG chart

DE-cr:
Tested with Firefox (Ubuntu), Chromium (Ubuntu), Chrome (Android), Silk (Android), Edge (Win11), Safari (Mac)
grafik
(It would have been much easier if the favicon was a light enough color right away: `.c3{background:url('favicon.svg')no-repeat center}`.)


fredlcore:
Nice :)!

PR #646, opened 2024-04-08
Include watermark in all html pages

DE-cr:
As an improvement (?) to https://github.com/fredlcore/BSB-LAN/pull/645, e.g.
Bildschirmfoto vom 2024-04-08 21-09-43
I would have preferred to have the logo stay at the same window position when scrolling, but unfortunately using `background-attachment:fixed` leads to the watermark being cut off when scrolling past the initial window's size. Using `height:200%` would have solved this (for /C and other long pages I tried), but then you could scroll down well past the content of pages that fit into the initial window.

Tested with the usual browsers (Ubuntu:Firefox,Chromium; Win11:Edge; Android:Chrome,Silk; Mac:Safari), at various window sizes.

### DIFF
--- a/BSB_LAN/html_strings.h
+++ b/BSB_LAN/html_strings.h
@@ -87,8 +87,6 @@ const char graph_html[] =
     ".c3-tooltip{opacity:0.7;background-color:#eee}" NEWLINE
     "th{background-color:#ccc}" NEWLINE
     ".value{text-align:right}" NEWLINE
-    // add BSB-LAN watermark (remove the following line if you don't like it):
-    ".c3:before{position:absolute;z-index:-1;width:100%;height:100%;content:'';opacity:.02;background:url('favicon.svg')no-repeat center}" NEWLINE
   "</style>" NEWLINE
   "<script src='" D3_LIBRARY_PATH "'></script>" NEWLINE
   "<script src='" C3_LIBRARY_PATH "'></script>" NEWLINE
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks, but I think in this case, it's sufficient to have the logo on the top of the page (otherwise it might seem the project is a bit too self-centered ;) ). With the graph, I have seen it here and there that people post it and some take a screenshot of the whole screen, but some just take one of the graph, so that's why the watermark here is really helpful. However what I was wondering whether it is possible to offer a "Download" button so that one could download a rendered version of the graph withouth having to do a screenshot?

DE-cr:
:D understood.

Graph download could be either SVG or bitmap (e.g. PNG). The former might be doable even without additional JavaScript libraries, but the files could become rather large with many data points rendered. The latter would probably require an additional library. Personally, I don't see why a bitmap download function would be worth the implementation, as screenshots are so easy to do (except on MacOS, where I haven't been able to memorize the key combination required to trigger one). An SVG download could be somewhat helpful, as it would allow zooming in on details. However, you'd still lose all the interactive functionality of the live plot - which you'd keep by doing a /D download and loading it in `scripts/BSB-LAN_datalog-viewer.html`. Btw, do you want a watermark included there as well?

fredlcore:
Yes, you're right, then we just keep it as it is.
PS: Screenshot on Mac is command+Shift+4 :)

PR #648, opened 2024-04-09
Simplify code in BSB-LAN_datalog-viewer.html

DE-cr:
Contrary to the BSB-LAN executable, we don't use an existing black svg here, but are creating our own one. Therefore we don't have to go through hoops to achieve opacity; we can just render the svg lighter.

Sorry for not thinking of this while https://github.com/fredlcore/BSB-LAN/pull/647 was still open.
Feel free to ignore this change proposal and just close the PR.

### DIFF
--- a/BSB_LAN/scripts/BSB-LAN_datalog-viewer.html
+++ b/BSB_LAN/scripts/BSB-LAN_datalog-viewer.html
@@ -29,9 +29,8 @@
       .c3-tooltip{opacity:0.7;background-color:#eee}
       .c3-tooltip th{background-color:#ccc}
       .c3-tooltip .value{text-align:right}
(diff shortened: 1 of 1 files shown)


fredlcore:
Sorry, but what does it do now, and why should I (not) merge it?

DE-cr:
It just makes the code easier.

You might want to discard it to
- avoid a merge that only makes the code of a rarely used (?) script easier to understand
- keep the code as close as possible to https://github.com/fredlcore/BSB-LAN/blob/master/BSB_LAN/html_strings.h (there, it has to be more complicated, see my initial comment here above)

fredlcore:
That's fine, thanks!

PR #665, opened 2024-09-04
/DG: added rescaling and subchart on zooming

DE-cr:
notes:
the subchart only becomes visible when zooming in on the chart
the subchart's gray area can be moved to move the zoom window
helpful subchart side effect: when zoomed in, de-/selecting curves doesn't zoom out, as it does w/o a subchart
to get rid of the subchart: reload /DG

before the code change:
Bildschirmfoto vom 2024-09-04 07-13-56

with the code change:
Bildschirmfoto vom 2024-09-04 07-13-44


fredlcore:
Nice, thank you!

PR #667, opened 2024-09-17
Update Web-Interface2.png: /DG screenshot with subchart

DE-cr:
To showcase that new functionality (added subchart when zoomed in).
Take it our leave it. ;)


fredlcore:
Thanks!

PR #676, opened 2024-11-28
Translate all the German parameters into English

plauwers:
I translated the file `lang_de.json` into `lang_en.json`, so that all the keys are now translated into english


fredlcore:
Thanks a lot!!

PR #679, opened 2024-11-29
Creation of german manual

plauwers:
AI generatin of a german manual


fredlcore:
Thanks a lot! Again, would it be possible to change to informal mode of address ("Du" instead of "Sie")? Because the verbs will change as well, this is not just a simple search and replace option, but if you could configure the AI model in such a way, that would be great.

plauwers:
I changed the prompt of the model, to create an informal version of the translation, all these changes are in the second commit

fredlcore:
Thanks a lot!

PR #718, opened 2025-01-26
Adapt the size of MQTT tempTopic to the size of config variable

uschindler:
Hi @fredlcore,

this is just something I found out while reviewing the code yesterday in PR #717: It adapts the buffersize for creating the topic (for listening on) in the same way like I have done yesterday for the hostname. As before, the sizeof is evaluated by the compiler, so compiled code is identical.

It is up to you to merge this although it fixes no real bug, but it makes the code more safe when modifying config file to larger topic buffer size and it is easier to read.

Uwe


fredlcore:
Good call, thanks!

PR #726, opened 2025-02-02
add measurement state class to some numeric parameters without unit

jstammi:
Adds the `state_class` for some types of parameters.

Most porbable my restriction does not meet anybodies needs already. It comprises only those types, that parameters in my custom definitions were affected by.


fredlcore:
As mentioned in the issue, this needs to be a general approach. Otherwise we end up listing various data types with which have no unit (of which there are dozens).

fredlcore:
I've come up with a more general approach in the most recent update.

PR #734, opened 2025-05-06
Update LANG_FR.h

Nerzh29:
Update LANG_FR and add link to LANG_FR_LEGACY.h


fredlcore:
Why do you re-translate entries such as MENU_TEXT_QIN which has already been translated into French? I already spent a lot of time going through the translations made by another user, so unless there is a really good reason why the previous translations are wrong, do not simply do redundant translations. 
If you really want to help improving the French translation, then first take a look at LANG_DE.h (not! LANG_DE_LEGACY.h!), check each entry if a corresponding entry in LANG_FR.h exists, and only add a new translation if one of the entries is missing in LANG_FR.h. Nothing else is required or accepted.

Nerzh29:
Okay, sorry. I'm not very familiar with how GitHub works. I will only modify the additional data.

fredlcore:
Thanks. This is not GitHub-related, by the way, it's the way BSB-LAN is structured. Since it has grown "organically" for over 10 years now, it's sometimes not always too obvious, so better ask first before you do a lot of work (for which I am, of course, grateful!)...

PR #735, opened 2025-05-06
Create LANG_FR_LEGACY.h

Nerzh29:
translate LANG_DE_LEGACY.h to LANG_FR_LEGACY.h


fredlcore:
As stated in LANG_DE.h at the bottom, LANG_DE_LEGACY.h is only necessary for the German language file as it contains the translations that nowadays come from the device-specific parameter list. This is only needed for old versions to compile correctly, and for those, no further support is provided.

PR #742, opened 2025-05-15
safe defaults for bsb bus pins on ESP32C3 and others

jsb1:
The default pin configuration can cause crashes on esp32c3 and maybe others.
Pins 16 and 17 have been defaults for any other processors than those of the predefined boards.
On esp32c3 these pins are flagged red in data sheet. They are used for internal communication to flash storage. Using them as gpio or uart is the safest way to crash the system ;-).

I added some better defaults for C3. Other unknown processors will end in a hint on debug output and an endless loop. I have no better idea, how to keep this change as small as possible.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -7599,9 +7599,16 @@ active_cmdtbl_size = sizeof(cmdtbl)/sizeof(cmdtbl[0]);
 #elif defined(ARDUINO_SAM_DUE)
       temp_bus_pins[0] = 19;          // RX2
       temp_bus_pins[1] = 18;          // TX2
+#elif (defined(CONFIG_IDF_TARGET_ESP32))
+      temp_bus_pins[0] = 16;          // NodeMCU ESP32 RX2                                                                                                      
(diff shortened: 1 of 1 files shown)


jsb1:
Hallo Frederik,

jetzt habe ich doch noch etwas gefunden, was unbedarfte User ins offene Messer laufen lässt. Das ist mir nicht passiert, weil ich dachte, man müsse die Bus-Pins ohnehin im config-header setzen. 

Mit dem Patch ist es ein klein wenig besser. Man sieht wenigstens, was zu tun ist.

Viele Grüße und Danke für diese Projekt,

Jörg

fredlcore:
Thank you for that, but I won't merge this. With hundreds of users using BSB-LAN, I have to focus my support on solutions that work out of the box as much as possible. The C3 isn't pin-compatible with the two board versions I provide, and once people see that the C3 is mentioned in the code, they'll expect it to work like the others when in fact it's not.
I'll probably set up a page in the Wiki with caveats for other microcontrollers and mention this issue, but as long as the C3 and other models so not have compelling features that warrant their specific support, I'll stick to the two established variants.

PR #757, opened 2025-09-17
Fix critical buffer overflow vulnerability in HTTP request parsing

jack-kerouac:
**Disclaimer**: When checking for security vulnerabilities, [Claude Code](https://claude.ai/code) found one for me. This PR fixes it. I let it walk me through the implications and the fix of this pull request. I am an absolute newbie in ESP32 programming and haven't written C in ages. Despite that, I considered opening this PR to be worthwhile, given that the problem description and fix that Claude Code explained to me sounded plausible. Following is the description of the actual problem and change.

---

This fixes a critical off-by-one buffer overflow in the HTTP request
handler that could lead to authentication bypass, information disclosure,
or crashes.

The vulnerability occurred in two places:
1. Line 4643: The buffer filling loop allowed writing up to index 251
   (252 bytes) without leaving room for null terminator
2. Line 4704: Null terminator was added without bounds checking, writing
   at index 252 which is out of bounds

Attack scenario:
- Sending exactly 252 bytes would overflow into bPlaceInBuffer variable
- This would reset the buffer index to 0 and create an unterminated string
- Subsequent string operations (strcpy, strncmp) would read beyond buffer
- Could bypass PASSKEY authentication or leak adjacent memory

Fix:
- Limit buffer filling to MaxArrayElement-1 to reserve space for null
- Add bounds checking before writing null terminator
- Ensure string is always properly null-terminated

This prevents any writes beyond the buffer bounds and eliminates the
authentication bypass vector.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -4640,7 +4640,7 @@ void loop() {
           }
         }
 
-        if ((c!='\n') && (c!='\r') && (bPlaceInBuffer<MaxArrayElement)) {
+        if ((c!='\n') && (c!='\r') && (bPlaceInBuffer<MaxArrayElement-1)) {
           cLineBuffer[bPlaceInBuffer++]=c;
           continue;
         }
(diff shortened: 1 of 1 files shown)


DE-cr:
P.s. Actually, I would just change this here:

`char  cLineBuffer[MaxArrayElement+1];  // line 4517: "+1" added`

jack-kerouac:
Thanks for the thorough review, @DE-cr! I've updated the PR to use the minimal fix you suggested.

Changes made:
1. Reverted the previous approach with bounds checking
2. Applied the simple +1 fix: char cLineBuffer[MaxArrayElement+1];
3. Bonus cleanup: removed the unnecessary post-increment operator when null-terminating

The array now has 253 bytes (indices 0-252), so the null terminator at index 252 is always valid.

Ready for another look when you have a chance!

fredlcore:
Sorry for the long delay and thanks for your work, both of you. PR merged.

PR #761, opened 2025-10-16
Esp32 lib bug workaround for /DG etc.

DE-cr:
One of the esp32's libraries currently has a bug that breaks /DG (and some other /D...'s) functionality.
This PR presents a workaround. See comments in the code and commits below.

Please note: As I have an esp32 NodeMcu only, I haven't been able to test the change on other bsb-lan configurations.


fredlcore:
Thanks for noticing this - it seems that the bug has been identified and fixed, so hopefully we can remove this once the next version of the framework is being released.

PR #801, opened 2026-03-14
Add GPIO input pull-up mode via /Gx,IP

acfischer42:
## Summary
- add GPIO input helper that supports pull-up mode
- keep existing  behavior ()
- add new  mode to configure 

## Motivation
Dry-contact inputs on ESP32 benefit from internal pull-up support 

## Implementation
- introduced `configureGpioInputMode(uint8_t pin, const char *dir_token)`
- in GPIO query path, replaced inline `INPUT` assignment with helper call
- helper handles:
  - `,I` => `INPUT`
  - `,IP` => `INPUT_PULLUP`


fredlcore:
Ok, but why create a function that is only used once? This makes reading/debugging the code harder if you have to jump around in the source code and there is no improvement in terms of other aspects.

PR #808, opened 2026-05-05
Handle missing CONFIG_FREERTOS_NUMBER_OF_CORES define for some build environments

TCWORLD:
When building with `PlatformIO`, this define does not appear to exist in the available libraries. I'm guessing related to which version of the Arduino ESP32 libraries are used for the build.

While trying to fix the build issue, I found an example on the Arduino community forum [here](https://forum.arduino.cc/t/esp32-c3-watch-dog-timer-wdt/1308760/2), in which the example for this watchdog initialisation uses `configNUM_CORES`. This define does exist in `FreeRTOSConfig.h` in the ESP32 libraries, and is defined in my case as 2 which matches the original hard-coded value before this define was used.

I wasn't sure of the source of the selected define, but clearly it exists somewhere in the Arduino IDE otherwise somebody would have noticed by now. So instead of just replacing it, I'm proposing we set a failback define in case the former is missing.

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -7801,6 +7801,10 @@ active_cmdtbl_size = sizeof(cmdtbl)/sizeof(cmdtbl[0]);
   //set watchdog timeout 150 seconds
     #define WDT_TIMEOUT 150
   #endif
+  #ifndef CONFIG_FREERTOS_NUMBER_OF_CORES
+  //Fail back to FreeRTOSConfig define for number of cores
+    #define CONFIG_FREERTOS_NUMBER_OF_CORES configNUM_CORES
+  #endif
 //    esp_task_wdt_init(WDT_TIMEOUT, true); //enable panic so ESP32 restarts
   esp_task_wdt_config_t config = {
     .timeout_ms = WDT_TIMEOUT * 1000,  //  60 seconds


fredlcore:
Hm, I'm using platformio (with pioarduino) as well, and the definement can be found in `.platformio/packages/framework-arduinoespressif32-libs/esp32/dio_qspi/include/sdkconfig.h`, where there are a lot of `CONFIG_FREERTOS_` definements. Which version of the arduino-esp32 framework are you using? I assume it's an older version because there are a number of users who use platformio, so the issue would have popped up earlier if it was a general issue.

TCWORLD:
I did wonder whether it was some versioning issue. I've been using:
 ```
platform = https://github.com/platformio/platform-espressif32.git
platform_packages = framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32.git#3.0.4, framework-arduinoespressif32-libs @ https://github.com/espressif/arduino-esp32/releases/download/3.0.4/esp32-arduino-libs-3.0.4.zip
```

It's possible I've missed it somewhere (did check the docs), but could you let me know what settings you have in your `platformio.ini` file and I'll update mine to match.

**Comment by fredlcore** [comment]
You're using version 3.0.4 which is quite outdated. BSB-LAN requires currently(!) at least 3.0.7. Besides, PlatformIO no longer supports Espressif/ESP32, so there is a community project (pioarduino) that has taken over. Here's my config for the Olimex EVB:

```
; PlatformIO Project Configuration File
;
;   Build options: build flags, source filter
;   Upload options: custom upload port, speed and extra flags
;   Library options: dependencies, extra library storages
;   Advanced options: extra scripting
;
; Please visit documentation for the other options and examples
; https://docs.platformio.org/page/projectconf.html

[platformio]
src_dir = src/BSB_LAN

[env:Olimex-EVB]
platform = https://github.com/pioarduino/platform-espressif32/releases/download/stable/platform-espressif32.zip
board = esp32-evb 
framework = arduino
build_src_filter = +<*>
upload_speed = 460800
board_build.partitions = min_spiffs.csv
build_flags = -DDS218 -DBSB_BUS -Wall -Werror=tautological-compare -Wtype-limits -Wimplicit-fallthrough
upload_port = /dev/cu.*usb*serial*
build_type = debug
monitor_port = socket://192.168.1.50:23
monitor_filters = esp32_exception_decoder, time
monitor_speed = 115200
```

TCWORLD:
Thanks. I'll update. Wasn't aware of `pioarduino`, so was using the most recent one on platformio that actually compiled at all.

Issue #809, opened 2026-05-06
[BUG] BSB-LAN disapear of MQTT / Home assistant


jakebslo:
**Before submitting a bug report, please make sure that you have checked the [FAQ](https://docs.bsb-lan.de/faq.html) and [troubleshooting section](https://docs.bsb-lan.de/troubleshooting.html) of the manual**. Most problems that are reported to us can be solved by going through the solutions provided there.

**Do not remove any of these questions! Bug-Reports that do not contain all required information will not be accepted. If you cannot provide any required information, explain why this is the case. Please write your bug report in English, BSB-LAN is used all across Europe, so writing in English allows other users to comment/benefit as well.**

**Version of the Arduino IDE**  
Arduino IDE 2.3.8

**Version of the ESP32 or Arduino framework**  
ESP32 3.3.8

**BSB-LAN Version**  
BSB-LAN, Version 5.1.6-20260425110220

**Architecture**  
ESP32 olimex EVB

**Bus system**  
BSB

**Describe the bug**  
When I perform an automatic MQTT discovery from the BSB-LAN settings, everything appears correctly in Home Assistant, including my approximately 900 sensors, but after a while—it can be minutes or hours—the BSB-LAN disappears from MQTT in Home Assistant.

**Log files - Bug reports without log files will be closed**  
If you cannot provide a log file, explain here why in order to prevent closing this bug report. Attach or copy/paste the log of the Arduino IDE's serial monitor when performing the actions that lead to the error. If possible start the log right before BSB-LAN boots (pressing the reset button on the board).

**Expected behavior**  

**Screenshots**  

**Additional context**  
Olimex installed next to the heat pump with a 5V/2A 40Hz-50Hz/100V-240V power supply adapter (European style plug).

LiYCY shielded control cable, 2 x 0.75 mm², between the heat pump and the Olimex.

DHCP is enabled but set to static from the internet box

[core_mosquitto_2026-05-06T09-14-39.119Z.log](https://github.com/user-attachments/files/27433081/core_mosquitto_2026-05-06T09-14-39.119Z.log)

fredlcore:
You are using a mosquitto version higher than 2.0.15. This is highlighted in our manual:

[...]

PR #810, opened 2026-05-08
Fix stack-buffer overflow in mqtt_callback for topic suffixes ≥ 10 chars

remidebette:
## Summary

`mqtt_callback()` has an unbounded `%s` in the fallback `sscanf` that writes into a 10-byte stack buffer. When BSB-LAN subscribes to `<MQTTTopicPrefix>/#` and the broker delivers a retained topic whose suffix after the prefix is 10 characters or longer (e.g. `BSB-LAN/0/38/10018`), the call writes past the end of `parsed_command[10]`, smashing the stack canary and triggering a panic on the next function epilogue.

## Reproduction

1. Configure BSB-LAN with MQTT enabled and `log_parameters` containing any 5-digit parameter ID together with a category whose digit count plus parameter ID makes the topic suffix ≥ 10 characters. Example: parameter `10018` in category `38` — suffix `/0/38/10018` is 11 bytes including the null terminator.
2. Let BSB-LAN publish at least once so the broker retains the message under that topic.
3. Reboot BSB-LAN. On reconnect, it subscribes to `<prefix>/#`, the broker replays the retained message, the callback parses it, and the device crashes:
   ```
   Stack smashing protect failure!
   Backtrace: 0x4008d9cc:0x3ffb1d80 ...
   ```
4. Because the message stays retained, the device enters a reboot loop on every restart until the topic is deleted from the broker or the bug is fixed.

## Root cause

In `BSB_LAN/include/mqtt_handler.h`, around line 593:

```c
char parsed_command[10];
...
} else if (sscanf(topic+topic_len, "/%s", parsed_command) == 1) {
```

The first `sscanf` a few lines above already uses the safe form (`%9s`); only this fallback branch was missed.

## Fix

```diff
-  } else if (sscanf(topic+topic_len, "/%s", parsed_command) == 1) {
+  } else if (sscanf(topic+topic_len, "/%9s", parsed_command) == 1) {
```

One-character change. No behavioral effect on valid commands (`poll`, `set`, `inf`, `status` are all ≤ 6 chars). Longer unrecognized suffixes are now safely truncated and still flow into the existing "Unknown command at the end of main MQTT topic" log line.

## Alternative considered

Bumping `parsed_command` to e.g. `char[32]` would also work, but `%9s` is the minimal, surgical fix that mirrors the bound already used in the sibling `sscanf` two lines above. If a larger buffer is preferred for future flexibility, both `sscanf` bounds and the array size should be updated together.

## Notes

- Discovered on a Joy-It ESP32 NodeMCU running BSB-LAN 5.1.1 with a Siemens RVS21.831/127 controller, but the bug is not hardware- or controller-specific — any retained MQTT message with a long enough topic suffix triggers it.
- Recovery without the fix requires deleting the offending retained topic from the broker (e.g. via MQTT Explorer → right-click → Delete) before the device can boot stably again.

### DIFF
--- a/BSB_LAN/include/mqtt_handler.h
+++ b/BSB_LAN/include/mqtt_handler.h
@@ -590,7 +590,7 @@ void mqtt_callback(char* topic, byte* passed_payload, unsigned int length) {
       return;
     }
   // Publish comma-separated parameters to /poll underneath main topic to update a number of parameters at once
-  } else if (sscanf(topic+topic_len, "/%s", parsed_command) == 1) {
+  } else if (sscanf(topic+topic_len, "/%9s", parsed_command) == 1) {
     if (!strcmp(parsed_command, "poll")) {
       printFmtToDebug("MQTT message received [%s | %s]\r\n", topic, payload);
       char* token;


fredlcore:
Thanks!

remidebette:
Related to #809

PR #813, opened 2026-05-12
Improve WiFi AP selection and fix reconnect issues on ESP32

schlaubstar:
This PR fixes unreliable WiFi roaming behavior on ESP32 devices and improves the logic that selects and connects to the strongest available access point. It also resolves an issue where defining a fixed BSSID caused cyclic reboots due to repeated reconnect attempts.

**Background / Original Issue**
The previous implementation did not reliably connect the ESP32 to the strongest access point. Even when stronger APs were available, the device often stayed connected to a weaker one.

To work around this, I tried defining a preferred BSSID. While this forced the device to connect to the intended AP, it introduced a new problem:
the device started rebooting in ~60‑second intervals, likely due to continuous WiFi reconnect attempts triggered by a bug in the existing logic.

**What I changed**
**1. Reworked** `scanAndConnectToStrongestNetwork()`
The function now handles the entire decision process on its own:

- Added a guard to only re‑scan APs when already connected but RSSI < –80 dBm
- Fixed a missing return that previously caused unintended disconnects()
- Ensured the function is self‑contained and no longer depends on external WiFi state assumptions

This makes the connection logic more predictable and prevents unnecessary reconnect cycles.

**2. Updated setup() logic**
Previously, setup() called WiFi.begin() before scanAndConnectToStrongestNetwork().
This caused the ESP32 to stick to weaker APs, because the initial connection was already established before the scan logic ran.

I removed this premature call so that the strongest‑AP selection happens before any connection attempt.

**Results / Testing**
The device now reliably connects to the strongest access point

Roaming between APs works as expected (tested switching to a weaker AP and back)

The logic no longer requires a fixed BSSID to maintain a stable connection

The BSSID‑based logic is still present, but I did not modify it — I suspect the reboot issue still exists there, but it is unrelated to the improvements in this PR

### DIFF
--- a/BSB_LAN/BSB_LAN.ino
+++ b/BSB_LAN/BSB_LAN.ino
@@ -7278,22 +7278,13 @@ next_parameter:
 
 #if defined(ESP32)
     if (network_type == WLAN) {
-  // if WiFi is down, try reconnecting every minute
-      bool not_preferred_bssid = false;
-      if (WiFi.BSSID() != nullptr) {
-        for (int x=0;x<6;x++) {
-          if (WiFi.BSSID()[x] != bssid[x] && bssid[x] > 0) {
-            not_preferred_bssid = true;
-          }
-        }
-      }
-  
-      if ((WiFi.status() != WL_CONNECTED || not_preferred_bssid == true) && localAP == false) {
-        printFmtToDebug("Reconnecting to WiFi...\r\n");
+      if (WiFi.status() != WL_CONNECTED && localAP == false) {
+        printFmtToDebug("WiFi not connected (status=%d), reconnecting...\r\n", WiFi.status());
         scanAndConnectToStrongestNetwork();
       }
     }
 #endif
+	  
   }
   if (bus->getBusDest() != dest_address) { // just in case temporary reset doesn't (always) work, reset detination ID back to default.
     printFmtToDebug("Current destination ID %d does not match default destination ID %d, fixing...\r\n", bus->getBusDest(), dest_address);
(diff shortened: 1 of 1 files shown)


fredlcore:
Thanks, but if there is a crash (probably due to some uncaught exceptions) related to this function, then this needs to be isolated/fixed first. Otherwise we change the code without knowing what causes this, and this creates even bigger problems.

schlaubstar:
Hi @fredlcore : After a lengthy debugging session using a serial connection, I need to revise my statement. The disconnects where not caused by a crash but happened "intentionally". 

My Unifi AP seems to change the BSSID if you change some configurations like channel/VLAN etc. 
That said .... my provided bssid in the config.h simply was out of date.

In the Serial output I could instantly see that the connection to the defined BSSID in the config was not possible - then the scan for strongest network started > connected - fine for a while. 
But on the next check it is detected that I we are not connected to the expected BSSID and then wifi is disconnected, scan started again etc. 
The ESP32 itself remains stable but I could not see this via Telnet because all "Wifi conection related output" is not visible in Telnet.

my proposed fix worked fine for all use-cases but as soon as you are connected with a good signal quality, it does not re-connect or care if there is a deviation between the provided bssid and the one actually connected to. I would propose to remove the option to provide a BSSID in general, as a working "reliably connect to the strongest" logic was what the users were looking for.

fredlcore:
Have you followed why the fixed BSSID was added to BSB-LAN? Or how would you know "what the users were looking for"? If you had submitted a bug report, you would have been forced to take a look at the serial monitor log, and it would have been immediately obvious that there is no crash in the first place. What you describe right now is exactly the functionality why the BSSID was implemented:
a) to pin BSB-LAN to a single router and prevent "hopping" in case there are two almost identical APs available, and 
b) to connect to "hidden" WiFi networks
In both cases, the users have chosen a specific AP by providing the BSSID. There is no reason to start ignoring that in some circumstances because there will always be other issues coming up then.

So, if your AP changes the BSSID due to different configurations, then this is rather unusual. In any case, this is not what BSB-LAN would/should take into account. Your initial problem would be solved if you choose the right BSSID and leave your router alone. Then it will provide to the strongest AP. And if that AP should, for some reason not be available, it will choose the strongest one and try again to reconnect after some time.

This is how it has worked (well) for years now, and I don't see a reason to fix something that isn't broken.

But thanks for your efforts, I appreciate users who are willing to contribute.

schlaubstar:
Hi @fredlcore : 
I observed that the ESP32 is not connecting with the stongest AP > That was the reason why I did manually set an BSSID .... which changed over time causing the reconnects.

Then I looked at the code first, instead of doing a serial debug and found this glitch:

Wifi.begin() in the Setup() connects you with anything that is strong or fast, ignoring a provided BSSID, independent if it is correct or wrong.

So you reach: 
scanAndConnectToStrongestNetwork().
- if you did not define a bssid > it will correctly connect you with the strongest AP, which is great.
- if you define a correct BSSID > it might find out "not_preferred_bssid == true" (as Wifi.begin() did not necessarily connect you with the right AP) and try to connect you but you are missing a return; in Line 7327 after you found out that you sucessfully connected to the right AP thus calling esp_wifi_disconnect(); and scanning for the strongest AP() without any necessity.

- if the provided BSSID does not belong to the strongest AP, you then end in the same "connect/reconnect" loop I was in, rendering "not_preferred_bssid == true" in each check, because after connecting to the right AP, you disconnect due to the missing return; and connect to the strongest AP again.

- if the provided BSSID is wrong, you anyway disconnect each time the cycle runs, which technically is ok because I here the fault was on my side  ;-)

So I would propose to add at least the missing return; in scanAndConnectToStrongestNetwork() after you found out: printlnToDebug("Connection successful using default BSSID.");

fredlcore:
You're right, thanks. As this was initially introduced to connect to hidden SSIDs, the problem did not come up because despite the additional, unnecessary scan, the connection eventually was established. With the now added `return;`, it should also work in your case. Can you please check the most recent version?

schlaubstar:
Checked with the current version of your repo and it works as expected. No reconnects.

fredlcore:
Great, thanks!

PR #819, opened 2026-07-17
Disable WiFi modem sleep to improve latency and prevent long-term packet loss on ESP32

waringer:
Disable WiFi modem sleep to reduce latency and packet loss. This increases power consumption slightly.


fredlcore:
Good idea, but this needs to be guarded by a `if (esp32_save_energy == false)` condition.

waringer:
Sure, i have changed the code, thanks for the hint :)

